PR #4, opened 2017-05-20
labelmaker

mwallman:
Labels GCE instances and attached disks


mwallman:
update email in author from mwallman@google to mwallman@google.com

PR #14, opened 2017-08-07
Figure redrawn in Google public format

mwallman:


woogie-wolgemuth:
The original image didn't have the header (Architecture: Cloud Connect) or footer (gray bar). Do you think those will look confusing in the context of https://github.com/GoogleCloudPlatform/professional-services/tree/master/infrastructure/cloudconnect ?

mwallman:
This change came from the SA (Miles) team.

PR #16, opened 2017-08-21
Spring Boot based Hello World app with Oauth2

hm-distro:


application-development/helloworld-springboot-auth/README.md
```diff
@@ -0,0 +1,106 @@
...
+   YOUR_PROJECT` to the ID of your application.
+1. Run `mvn spring-boot:run`
+1. Visit http://localhost:8080
+
+
+## Deploy to App Engine flexible environment
+
+1. `mvn appengine:deploy`
+1. Visit `http://YOUR_PROJECT.appspot.com`.
+* if this is the first deployment, you need to run `gcloud app create`
```
joseret: This needs to be clearer and I would add to the instructions to make sure

application-development/helloworld-springboot-auth/src/main/java/com/example/java/gettingstarted/WebSecurityConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.example.java.gettingstarted;
+
+import org.springframework.boot.autoconfigure.security.oauth2.client.EnableOAuth2Sso;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+@EnableOAuth2Sso
+@Configuration
+public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
```
joseret: Would show examples of protected and unprotected to be even more useful as to what goes here (unless you are assuming the rest will be in spring documentation).  At least I would put a link.

application-development/helloworld-springboot-auth/README.md
hm-distro: Which part? The first deployment ?
I do have links  to the documentation in the start of the README

+[App Engine-flexible]: https://cloud.google.com/appengine/docs/flexible/
 +[cloud-java]: https://cloud.google.com/java/
 +[spring-boot]: http://projects.spring.io/spring-boot/

application-development/helloworld-springboot-auth/src/main/java/com/example/java/gettingstarted/WebSecurityConfig.java
hm-distro: This is based off https://github.com/GoogleCloudPlatform/getting-started-java/tree/master/helloworld-springboot , if one wants to  try out the unprotected end points.

application-development/helloworld-springboot-auth/README.md
joseret: Yes, the first deployment.  The error that mvn gives is hidden and hard to find.  Simple change to avoid a bad experience :)

application-development/helloworld-springboot-auth/src/main/java/com/example/java/gettingstarted/WebSecurityConfig.java
joseret: I understand, so at least give the links, but again you are forcing the user to go to another place.  I would prefer to see the sample have some of this and then still have the link(s) to understand this deeper.  Just a suggestion.

hm-distro: https://github.com/GoogleCloudPlatform/professional-services/pull/20 has these changes. Thanks

PR #22, opened 2017-12-04
Remove_dag.py : a dag to remove dags

smeyn:
This is a contribution to the Composer (Apache Airflow) project. It implements a simple DAG that allows the invoker to remove another DAG in the composer database. 
This addresses the fact that Composer at the moment does not have a UI to remove DAGs


ryanmcdowell:
Closing as there are no changes in this PR.

PR #29, opened 2018-01-08
resources inventory - list disks

hm-distro:
Initial version of script to list the disk resources


hm-distro:
Closing as newer  version is on the way

PR #32, opened 2018-02-15
Feature/remove public gcs cf

reechar-goog:


infrastructure/removePublicGCS/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Begin by activating a *Cloud Shell* within your [Cloud Console](https://console.cloud.google.com/).
+
+    ## Make a new directory as a staging area to deploy the cloud function
+    $ mkdir removePublicAccessGCF
+    $ cd removePublicAccessGCF
+
+
+    ## Grab a copy of the sample Cloud Function code
+    $ wget https://raw.githubusercontent.com/reechar-goog/professional-services/feature/removePublicGCS_CF/infrastructure/removePublicGCS/index.js
```
woogie-wolgemuth: Switch these instructions to point at the PSO GitHub; also, prefer using Git clone over wget raw files (we can assume git competency for visitors to GitHub).

infrastructure/removePublicGCS/README.md
```diff
@@ -0,0 +1,34 @@
...
+    $ wget https://raw.githubusercontent.com/reechar-goog/professional-services/feature/removePublicGCS_CF/infrastructure/removePublicGCS/package.json
+
+
+    ## Use NPM to install a copy of the dependencies 
+    $ npm install --save @google-cloud/storage
```
woogie-wolgemuth: Switch the command to rely on your package.json to list the dependencies instead.

infrastructure/removePublicGCS/index.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {function} callback The callback function.
+ */
+
+const gcs = require('@google-cloud/storage')();
+
+exports.removePublicAccess = function (event, callback) {
+  const object = event.data;
+  const file = gcs.bucket(object.bucket).file(object.name);
+  console.log(`Ensuring no public ACLs on ${file.name}`);
+  file.acl.owners.deleteAllAuthenticatedUsers();
```
woogie-wolgemuth: This pattern needs a few elements of cleanup.

* The deleteFoo() methods are blind - each of those will issue an API call whether or not the ACL includes the unwanted users. In the `event` object you should be receiving the resource representation of the Object, including the ACLS (https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations). You can greatly reduce the relative cost of this function by checking the ACLs for the presence of 'allUsers' or 'allAuthenticatedUsers' before invoking.
* Each of these function calls is returning a Promise since there's no callback provided, but you're dropping the Promise. You should issue your API calls in parallel like this, but wait for them all to return - see examples here: https://coderwall.com/p/ijy61g/promise-chains-with-node-js
* You need to have some error logging in the event your call to modify the ACL doesn't work - see above about checking for the results of the promises from your API calls.

infrastructure/removePublicGCS/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@google-cloud/storage": "*"
```
woogie-wolgemuth: Cloud functions are pretty forgiving because you don't have to worry about having a complicated environment for dependencies, but it's still a good idea to be a little more restrictive here than '*' for versioning. I'd suggest setting this to 'at least version I tested with' and 'less than next API breaking version', like so:
>=1.5.2 <2

infrastructure/removePublicGCS/README.md
```diff
@@ -0,0 +1,34 @@
+# removePublicGCS - Sample 5-minute Solution to prevent public objects in GCS via Cloud Functions
+
+This is a sample of deploying a *Cloud Function* that triggers off a target *GCS* bucket whenever an object's metadata or ACLs are updated. For each object that is updated, this function will explicitly remove the **'allAuthenticatedUsers'** and **'allUsers'** special ACL groups from the owners, readers, and writers groups. This function effectively reacts and removes public access if it is ever accidentally granted by a user or code. The object getting the public ACLs removed are not modified in any other way, the object data, metadata, and other ACLs will remain the same after the function triggers.
+
+## Steps
```
woogie-wolgemuth: You'll need something in your steps that calls out our docs about Service Accounts for Cloud Functions, as well as the GCS roles that the SA will need to do this work. Including a gcloud command that tests the IAM permissions for modifying an object on the target bucket would really help with troubleshooting as well.

infrastructure/removePublicGCS/package.json
woogie-wolgemuth: GitHub ate my damn comment, >=1.5.2 <2

ryanmcdowell:
Closing as this is > 6 months old. Please re-open and complete review if still valid.

PR #34, opened 2018-05-10
BigQuery cross project slots utilization monitoring

omerhabas:
AppEngine application that solves monitoring of BigQuery slot utilization across projects, as follows:
- Billing API is used to dynamically and automatically discover projects using the billing account ID in hand.
- Monitoring API is used to create a custom metric in a defined central Stackdriver account.
- Monitoring API is used to read metrics from all linked projects, and write to the custom metric created in the central Stackdriver account. This removes the need of linking all projects to a Stackdriver.
- Datastore is used to record last fetched data point from each project. This is needed to only read latest data from each project.


omerhabas:
CLA issue due to author / committer mail issue.

PR #38, opened 2018-06-06
Added cloud-composer-examples.

sabhyankar:
Added an example Cloud Composer workflow that demonstrates how a Dataflow job can be triggered to process, transform and load data into a BigQuery table. The workflow is triggered via Cloud Function.


data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
```diff
@@ -0,0 +1,136 @@
...
+        self.filename_dict = {'filename': filename}
+        self.load_dt_dict = {'load_dt': load_dt}
+
+    def parse(self, row):
+        """This method translates a single delimited record into a dictionary
+        which can be loaded into BigQuery. It also adds filename and load_dt
+        fields to the dictionary."""
+
+        # Strip out the return characters and quote characters.
+        values = re.split(self.delimiter, re.sub('\r\n', '', re.sub(u'"', '', row)))
```
dparrish: You can do this in a single re.sub call to avoid creating an extra string object:

`re.sub(r'[\r\n"]', '', row)`

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
```diff
@@ -0,0 +1,136 @@
...
+
+        row = dict(zip(self.keys, values))
+
+        # Add an additional filename field.
```
dparrish: This is an extremely inefficient way of adding these fields, it results in multiple copies being made of every row, and on a large input file that can really be substantial.

Instead, you don't need to create a `self.filename_dict` and `self.load_dt_dict`, just save the provided `filename` and `load_dt` in the constructor, then here:

```
row['filename'] = self.filename
row['load_dt'] = self.load_dt
return row
```

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
```diff
@@ -0,0 +1,136 @@
...
+import argparse
+import logging
+import ntpath
+import re
+
+import apache_beam as beam
+from apache_beam.options import pipeline_options
+
+
+class RowTransformer(object):
```
dparrish: Can this not be done with the built-in csv class?

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
```diff
@@ -0,0 +1,136 @@
...
+    """A helper class that contains utility methods to parse the delimited file
+    and convert every record into a format acceptable to BigQuery. It also contains
+    utility methods to add a load_dt and a filename fields as demonstration of
+    how records can be enriched as part of the load process."""
+
+    def __init__(self, delimiter, header, filename, load_dt):
+        self.delimiter = delimiter
+
+        # Extract the field name keys from the comma separated input.
+        self.keys = re.split(',', header)
```
dparrish: Is the header always comma-separated, even when the delimiter is not?

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
```diff
@@ -0,0 +1,136 @@
...
+                        default=',')
+
+    parser.add_argument('--fields', dest='fields', required=True,
+                        help='Comma separated list of field names.')
+
+    parser.add_argument('--load_dt', dest='load_dt', required=True,
+                        help='Load date in YYYY-MM-DD format.')
+
+    known_args, pipeline_args = parser.parse_known_args(argv)
+    parse_delimited = RowTransformer(delimiter=known_args.delimiter,
```
dparrish: You have to provide the header as an option to this script, even though the header is in the input file? I can see that getting out of sync.

data-analytics/cloud-composer-examples/composer_dataflow_examples/simple_load_dag.py
```diff
@@ -0,0 +1,147 @@
...
+    # the information about the GCS object that triggered this DAG.
+    # We extract the bucket and object name from this dictionary.
+    source_bucket = kwargs['dag_run'].conf['bucket']
+    source_object = kwargs['dag_run'].conf['name']
+    completion_ds = kwargs['ds']
+
+    target_object = os.path.join(target_infix, completion_ds, source_object)
+
+    logging.info('Coping %s to %s',
+                 ''.join([source_bucket, '/', source_object]),
```
dparrish: `'/'.join((source_bucket, source_object))`

or 

`os.path.join(source_bucket, source_object)`

data-analytics/cloud-composer-examples/tests/test_cloud_composer_examples.py
```diff
@@ -0,0 +1,54 @@
...
+                                         load_dt=test_load_dt)
+        csv_input = 'KS,F,1923,Dorothy,654,11/28/2016'
+
+        expected_dict_output = {u'state': u'KS',
+                                u'gender': u'F',
+                                u'year': u'1923',
+                                u'name': u'Dorothy',
+                                u'number': u'654',
+                                u'created_date': u'11/28/2016',
+                                u'filename': test_filename,
```
dparrish: You don't need the `u` prefix on the dictionary keys.

In fact I'd say this is wrong because `filename` and `load_dt` are set _without_ the unicode prefix in RowTransformer, so you shouldn't be requiring it here.

sabhyankar:
Hey David - Thanks for the super quick review. I have completed all the changes you had recommended and have put some additional comments above. Let me know if you see something else. Thx again.

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
sabhyankar: Agreed! Since this is meant to be an example, I have removed the header from the sample file and only use the header provided as an input.

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
sabhyankar: That is correct. The header in the example is always a comma separated input added as an Airflow variable. I have described that in the README.

data-analytics/cloud-composer-examples/composer_dataflow_examples/dataflow/process_delimited.py
sabhyankar: Yes it can. However, the intent is for customers to take this simple approach/example class and replace this with transformations with more complex business logic.

data-analytics/cloud-composer-examples/composer_dataflow_examples/simple_load_dag.py
```diff
@@ -0,0 +1,147 @@
...
+#                  Dataflow in <dataset>.<tablename> format.
+#                  e.g. lake.usa_names
+# input_field_names: Comma separated field names for the delimited input file.
+#                  e.g. state,gender,year,name,number,created_date
+with models.DAG(dag_id='GcsToBigQueryTriggered',
+                description='A DAG triggered by an external Cloud Function',
+                schedule_interval=None, default_args=DEFAULT_DAG_ARGS) as dag:
+    # Args required for the Dataflow job.
+    job_args = {
+        'input': 'gs://{{ dag_run.conf.get("bucket") }}/{{ dag_run.conf.get("name") }}',
```
jacobferriero: Would it be better here to use `'input': 'gs://{{ dag_run.conf["bucket"] }}/{{ dag_run.conf["name"] }}',` to give the user a KeyError  if the `dag_run.conf` doesn't contain these keys? For example, if the user manually triggers this DAG from the UI (therefore the `dag_run.conf` would not have the `"bucket"` or `"name"` keys) then they would kick off a DataFlow job that is DOA. I could see using the `.get` method if your wanted to add a default value that might lead to a successful DataFlow job but it seems we should error out before spinning up DataFlow workers and waiting for DataFlow to handle the error.

data-analytics/cloud-composer-examples/composer_dataflow_examples/simple_load_dag.py
```diff
@@ -0,0 +1,147 @@
...
+    # provides a dag_run.conf dictionary with event attributes that specify
+    # the information about the GCS object that triggered this DAG.
+    # We extract the bucket and object name from this dictionary.
+    source_bucket = kwargs['dag_run'].conf['bucket']
+    source_object = kwargs['dag_run'].conf['name']
+    completion_ds = kwargs['ds']
+
+    target_object = os.path.join(target_infix, completion_ds, source_object)
+
+    logging.info('Coping %s to %s',
```
jacobferriero: "Coping" do you mean "Copying"?

PR #73, opened 2018-08-27
Add GKE Billing Export

dparrish:
Add the GKE Billing Export tool, ariane id 259597


infrastructure/gke-billing-export/billing.go
```diff
@@ -0,0 +1,345 @@
...
+	clusters := getAllClusters(ctx, config)
+	config.AddValidator(func(old, new *autoconfig.Config) error {
+		clusters = getAllClusters(ctx, new)
+		return nil
+	})
+
+	// Start all background goroutines.
+	ch := make(chan *Cluster, *numGoroutines)
+	for i := 0; i < *numGoroutines; i++ {
+		go func() {
```
jasonmar: This looks like a pretty large function to nest, it would help readability a lot if you could create the function with another function which accepts the channel as an argument and returns func. This would also make it clear which variables are used from the outer scope.

`func getMetricsFunc(ch chan *Cluster) func { ... }`

```
getMetrics := getMetricsFunc(ch)
go getMetrics()
```

dparrish: I've moved it into a separate function. I don't think there's a need for the function indirection though.

infrastructure/gke-billing-export/README.md
```diff
@@ -0,0 +1,124 @@
+# gke-billing-export
+
+Google Kubernetes Engine fine grained billing export. This consists of a single binary that can be run in a Google Cloud project, which with appropriate permissions can scrape the metrics server for many Kubernetes Engine clusters. The metrics are stored in a dedicated BigQuery dataset.
+
+This is not an officially supported Google product.
+
+# Example Output
```
jasonmar: This is potentially confusing, because it is showing output from the bq query tool rather than 'billing'.
Can you add a note saying this is data that was inserted by the billing tool?

infrastructure/gke-billing-export/README.md
```diff
@@ -0,0 +1,124 @@
+# gke-billing-export
+
+Google Kubernetes Engine fine grained billing export. This consists of a single binary that can be run in a Google Cloud project, which with appropriate permissions can scrape the metrics server for many Kubernetes Engine clusters. The metrics are stored in a dedicated BigQuery dataset.
+
```
jasonmar: Things I'd suggest adding to the readme to help users understand what's provided:
- what's included in the config file?
- mention that the binary is `billing` since this is different from go default of using directory name
- how the tool is built with docker and deployed with deployment manager

Here's a starting point:

The executable takes no arguments, it reads project id and polling interval from  [config.json](config.json) and runs continuously as a daemon.
This repository provides a docker file for creating an image with the executable baked in, as well as a deployment manager configuration to run the image on your GKE cluster.
Once deployed, the billing export daemon will periodically poll the metrics API and save results to BigQuery.
```

infrastructure/gke-billing-export/README.md
dparrish: Done.

infrastructure/gke-billing-export/README.md
dparrish: Done, thanks!

jasonmar:
Verified that the program compiles and executes on my machine

PR #78, opened 2018-09-24
Restructure repository

ryanmcdowell:
Restructure repository to remove practice areas and provide a more simplified browsing experience.


ryanmcdowell:
I would agree that fake generator belongs under tools. However, I'm unsure of it's dependencies so I punted as to not break it and allow Jake or Josh to move and validate it.

For Medium posts, we can always update the links in the post provided we know what posts will be effected.

PR #79, opened 2018-09-27
Adding a tool to sync site verification (for GCS buckets) and a Google Group

jamesduke:
A tool to provision site verification (in the Seach Console) based on Google Group membership


jamesduke:
Trigger CLA check again

PR #84, opened 2018-09-28
Add Dataflow Streaming Benchmark Utility

ryanmcdowell:
This PR adds a Dataflow pipeline which publishes randomized fake JSON messages to a Cloud Pub/Sub topic at a configured QPS.


examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+    @ProcessElement
+    public void processElement(ProcessContext context)
+        throws IOException, JsonDataGeneratorException {
+
+      byte[] payload;
+      Map<String, String> attributes = Maps.newHashMap();
+
+      // Generate the fake JSON according to the schema.
+      try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+        JsonDataGeneratorImpl parser = new JsonDataGeneratorImpl();
```
sabhyankar: We can reuse the parser instead of creating a new parser per trigger. You can initialize the parser in the setup().

sabhyankar: This can be JsonDataGenerator interface.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * The main entry-point for pipeline execution. This method will start the pipeline but will not
+   * wait for it's execution to finish. If blocking execution is required, use the {@link
+   * StreamingBenchmark#run(Options)} method to start the pipeline and invoke {@code
+   * result.waitUntilFinish()} on the {@link PipelineResult}.
+   *
+   * @param args The command-line args passed by the executor.
+   */
+  public static void main(String[] args) {
+    Options options = PipelineOptionsFactory.fromArgs(args).as(Options.class);
```
sabhyankar: Add .withValidation() after you annotate the options fields with Validation.Required

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+
+      // Generate the fake JSON according to the schema.
+      try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+        JsonDataGeneratorImpl parser = new JsonDataGeneratorImpl();
+        parser.generateTestDataJson(schema, byteArrayOutputStream);
+
+        payload = byteArrayOutputStream.toByteArray();
+      }
+
+      context.output(new PubsubMessage(payload, attributes));
```
sabhyankar: Would it make sense to just publish json strings instead of a PubsubMessage object? We are not really passing any attributes so it might be easier for the downstream system to just consume UTF-8 encoded strings vs consuming PubsubMessage object.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+ * --topic=TOPIC"
+ * </pre>
+ */
+public class StreamingBenchmark {
+
+  /**
+   * The {@link Options} class provides the custom execution options passed by the executor at the
+   * command-line.
+   */
+  public interface Options extends PipelineOptions {
```
sabhyankar: Can this be private?

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+    @Description("The QPS which the benchmark should output to Pub/Sub.")
+    Long getQps();
```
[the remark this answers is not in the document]
ryanmcdowell: Added.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
ryanmcdowell: Eventually I imagine we can add functionality whereby the user can specify the eventId and eventTimestamp fields and we can populate those in the attributes. Added a TODO.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
ryanmcdowell: I normally leave Options public so that someone can programmatic-ly invoke the class by constructing the Options object and passing them to the run() method.

examples/dataflow-streaming-benchmark/README.md
```diff
@@ -0,0 +1,100 @@
...
+generates fake JSON messages matching the schema to a Pub/Sub topic at the rate of the QPS.
+
+## Pipeline
+
+[StreamingBenchmark](src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java) -
+A streaming pipeline which generates messages at a specified rate to a Pub/Sub topic. The messages 
+are generated according to a schema template which instructs the pipeline how to populate the 
+messages with fake data compliant to constraints.
+
+> Note the number of workers executing the pipeline must be large enough to support the supplied
```
jamesduke: This could likely trip up new users. Can we check the options in the main class to check that we have sufficient cores for the specified QPS?

ryanmcdowell: Looked into this. It doesn't appear the juice is worth the squeeze. Checking the options is failure prone because they are reliant on machineType which isn't validated until the job is submitted to the cloud. This adds a whole bunch of failure cases. For now, I would say let's merge this as folks are waiting on it and I can continue to look into if there's a cleaner way to accomplish what you're asking for.

PR #92, opened 2018-10-11
Initial import of gce-quota-sync

ludoo:
Asking for review and hopefully inclusion of this small tool, that syncs (global, regional) quotas from the GCE API to Stackdriver as custom metrics. Its purpose is of course to then be able to set alert policies in Stackdriver on specific quota usage.

The tool can either be run standalone, or deployed as a container to GKE/containers on GCE using the included Dockerfile.


tools/gce-quota-sync/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+  /usr/bin/pip install -r requirements.txt && \
+  apk del build-base python-dev && \
+  rm /var/cache/apk/*
+
+COPY gce-quota-sync.py /
+RUN chmod 755 /gce-quota-sync.py
+
+CMD \
+  echo "$CRONSPEC	/gce-quota-sync.py >>$LOGFILE 2>&1" > $CRONTAB && \
+  echo "3 0 * * * echo '' >$LOGFILE" >> $CRONTAB && \
```
woogie-wolgemuth: Why not have the crontab file be a simple file you overwrite when building the container?

tools/gce-quota-sync/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+* interfacing with the GCE API, and calling the
+  [`projects.get`](https://cloud.google.com/compute/docs/reference/rest/v1/projects/get)
+  and
+  [`regions.get`](https://cloud.google.com/compute/docs/reference/rest/v1/regions/get)
+  methods for global and regional quotas
+* deriving current utilization (usage / limit) for each quota, then converting
+  it into a time series
+* writing the time series to Stackdriver in the
+  `custom.googleapis.com/quota/gce` metric
```
woogie-wolgemuth: Terminating period.

tools/gce-quota-sync/requirements.txt
```diff
@@ -0,0 +1,4 @@
+click
```
woogie-wolgemuth: Could we pin the requirements to a least a major version revision?

tools/gce-quota-sync/README.md
```diff
@@ -0,0 +1,116 @@
...
+* `project`, eg `my-project-id`
+* `region`, eg `europe-west3`
+* `metric`, the actual quota value eg `CPU`
+* `limit`, the quota limit
+* `usage`, the quota usage
+
+The metric can then be used in charts, but more importantly to set custom
+alerts so as to monitor actual consumption, like in the following image:
+
+![](./images/quota_alert.png "Quota alert on threshold")
```
woogie-wolgemuth: Please provide details in the alt text, like "shows the Stackdriver UI, where a quota alert has been generated because a monitored metric has exceeded the threshold".

tools/gce-quota-sync/gce-quota-sync.py
```diff
@@ -0,0 +1,177 @@
...
+@click.option('--verbose', is_flag=True, help='Verbose output')
+def main(project=None, gce_regions=None, verbose=False, **kw):
+  "Fetch, convert, and write quotas for project and optional regions."
+  _configure_logging(verbose=verbose)
+  gce_regions = ['global'] + [s.strip() for s in gce_regions.split(',')]
+  quotas = []
+  try:
+    compute = googleapiclient.discovery.build(
+        'compute', 'v1', cache_discovery=False)
+    # fetch quotas for global + defined regions
```
woogie-wolgemuth: Capitalize first letter, period.

tools/gce-quota-sync/gce-quota-sync.py
```diff
@@ -0,0 +1,177 @@
...
+    for region in gce_regions:
+      _LOGGER.debug('fetching project quota for %s %s', project, region)
+      for quota in _fetch_quotas(project, region, compute=compute):
+        quotas.append((region, quota))
+    # convert quotas to series, write to Stackdriver using poor man's batching
```
woogie-wolgemuth: Same.

woogie-wolgemuth: Also, maybe 'naive' or 'simple' batching.

tools/gce-quota-sync/sample-deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
```
woogie-wolgemuth: Have you considered making this a simple Helm chart instead? (https://docs.helm.sh/chart_template_guide/)
Even in a small deployment, it can be hard to determine what values are variables that should be altered by the end user.

tools/gce-quota-sync/README.md
```diff
@@ -0,0 +1,116 @@
...
+  and
+  [`regions.get`](https://cloud.google.com/compute/docs/reference/rest/v1/regions/get)
+  methods for global and regional quotas
+* deriving current utilization (usage / limit) for each quota, then converting
+  it into a time series
+* writing the time series to Stackdriver in the
+  `custom.googleapis.com/quota/gce` metric
+
+
+## Usage
```
woogie-wolgemuth: You should mention which versions of Python are supported, e.g. 2.7 and 3.4+, only 2.7, only 3.5+, etc.

tools/gce-quota-sync/Dockerfile
ludoo: Because you might want to tweak the interval without rebuilding. This makes it very easy to test different schedules, not very important in this context but still practical, especially if the people running the deployment are != from the people building and pushing containers.

tools/gce-quota-sync/sample-deployment.yaml
ludoo: The deployment is so simple that a Helm chart looks a bit overkill (and not everybody uses Helm). One other option might be to remove the sample deployment file, and show the environment block in the README.

woogie-wolgemuth: Yeah, removing the deployment looks to be the way to go. You don't have to go all-in with Helm+Tiller, you can just do standalone deployments, but they don't advertise that fact very extensively.

ludoo:
Addressed all comments (thanks!) except two: one on the crontab variable/file, where my strong preference is to keep it in a variable for ease of tweaking without having to rebuild the container; and a second on the Helm chart, which I would skip (and maybe scrap the sample deployment, and just show the vars snippet in the README).

Let me know what you think!

tools/gce-quota-sync/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM alpine:3.8
+
+ENV CRONSPEC "0 * * * *"
```
woogie-wolgemuth: One last thought - if we're providing cron defaults, we should avoid 'crowded' slots like top of the hour. Throw a favourite prime number from 1 to 59 in the minute column.

tools/gce-quota-sync/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Copyright 2017 Google Inc.
```
ryanmcdowell: Why not 2018?

tools/gce-quota-sync/gce-quota-sync.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Google Inc.
```
ryanmcdowell: Why not 2018?

ludoo: Gah stupid me, copied it from source files I had around... thanks for noticing!

README.md
```diff
@@ -23,6 +23,7 @@ The tools folder contains ready-made utilities which can simpilfy Google Cloud P
 * [GKE Billing Export](/tools/gke-billing-export) - Google Kubernetes Engine fine grained billing export.
 * [LabelMaker](/tools/labelmaker) - A tool that reads key:value pairs from a json file and labels the running instance and all attached drives accordingly.
 * [Site Verification Group Sync](/tools/site-verification-group-sync) - A tool to provision "verified owner" permissions (to create GCS buckets with custom dns) based on membership of a Google Group.
+* [GCE Quota Sync](/tools/gce-quota-sync) - A tool that fetches resource quota usage from the GCE API and synchronizes it to Stackdriver as a custom metric, where it can be used to define automated alerts.
```
ryanmcdowell: Alphabetic order

ludoo:
Thanks! Hyper-quick review and merge.

PR #98, opened 2018-10-17
Adding the GCS Bucket Mover tool.

Tho77:
A prototype command line tool for moving a GCS bucket between projects, or within the same project to change settings such as location or default storage class.


Tho77:
Apparently need to redo everything with my google.com email address.

PR #106, opened 2018-11-06
TF HTTP/S, TCP GLB module

apollo-3:
A TF module that can create HTTP and/or HTTPS GLB as well as TCP GLB. In comparison with other similar modules this one solves named ports issue, supports TCP GLB and is more configurable in general.


morgante:
@apollo-3 Have you seen these modules?

- https://github.com/GoogleCloudPlatform/terraform-google-lb
- https://github.com/GoogleCloudPlatform/terraform-google-lb-http

If there are improvements for those, @danisla is generally very open to pull requests. Happy to work with you on those.

apollo-3:
@morgante yes I'v seen them, but I wouldn't say that our module is an improvement. It's more like a different module from those 2 above. Our module supports HTTP and/or HTTPS GLB and TCP GLB. In my opinion PR to those 2 repos doesn't make much sense. Instead a separate repo needed with the name something like "terraform-glb"
@danisla do you think it's possible to create a new repo for out TF module?

morgante:
@apollo-3 I'm not sure I follow. We already have modules for HTTP(S) GLB and TCP GLB so what is the difference with this module?

We shouldn't be propagating additional nearly-identical modules.

apollo-3:
@morgante, the differences are
1. Our module adds named ports
2. HTTP and HTTPS can be created separately or together
3. We were asked to combine HTTP and TCP GLB in one module
4. For our use case unneeded firewall part was removed and certificate management simplified
5. As the result of all above it’s a completely different module with a different variable set. Probably there are other differences that I can’t recall now

PR #111, opened 2018-11-10
Reorganizing Data Generator for BigQuery

jaketf:
There are several updates coming to the data generator this PR reorganizes the code for the coming updates. Here is a high level descriptions of the changes:

- Reorganize the code to encapsulate the same DataGenerator class containing data generation logic for demos to be used by several pipelines.
- Unpin the dataflow-python-examples dependencies in requirements.txt.
- Add feature to have numeric columns demonstrate an positive trend over the time range.
- Add feature to generate data with parmeterized key skew.
- Add bigquery script to copy a table into itself until it reaches a desired size (when time is of the essence).

This lays the ground work for the next PR which will include a companion pipeline with the ability to generate tables with joinable schemas.


ryanmcdowell:
Given that the data generator has grown quite a bit... does it make sense to pull it out of dataflow-python-examples into it's own top-level tool?

jaketf:
@ryanmcdowell Good idea. I can move it to a top-level tool in the next PR (when I will add another beam pipeline for generating a dimension table from an existing fact table based on the keys). I have a few updates in my backlog to get pushed out publicly. In the interest of having smaller incremental PRs can we move ahead with review of this PR as is?

jaketf:
going to submit a separate PR for pulling data generator out as top level example

PR #113, opened 2018-11-15
Update ephemeral_dataproc_spark_dag.py

jaketf:
Qwiklab blocked by exceeding vCPUs quota.


jaketf:
Reverted two commits that should not have been part of this PR

PR #134, opened 2018-12-28
[DONOT MERGE WIP]Configurable lock + Stackdriver + Tox

eshen1991:
**Configurable Lock**

**Stackdriver**

**Packaging with Tox**
For ease of use, a Tox ini file was added to enable packaging the tool into a single Pex file that can be run without needing to download any requirements or set up any virtual environments.In order to be able to make a .pex package from MacOS that will run on Linux, a special PyYAML egg is included that is compiled without the OS specific libyaml package.The resulting .pex package is in the /package folder.
To support packaging, all environment variables/global testing variables were refactored to be specified in a config yaml file or the command line.


eshen1991:
Got too many conflicts to resolve. close this PR and will open another one.

PR #135, opened 2018-12-28
Configurable lock + Stackdriver + Tox

eshen1991:
**Configurable Lock** 

**Stackdriver** 

**Packaging with Tox** 
For ease of use, a Tox ini file was added to enable packaging the tool into a single Pex file that can be run without needing to download any requirements or set up any virtual environments.In order to be able to make a .pex package from MacOS that will run on Linux, a special PyYAML egg is included that is compiled without the OS specific libyaml package.The resulting .pex package is in the /package folder. To support packaging, all environment variables/global testing variables were refactored to be specified in a config yaml file or the command line.


tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -33,48 +32,71 @@
...
+        conf: the configargparser parsing of command line options
     """
 
     # Load the environment config values set in config.sh and create the storage clients.
     config = configuration.Configuration.from_conf(conf)
 
-    print('Using the following service accounts for GCS credentials: ')
-    print('Source Project - {}'.format(
+    # We need to create the cloud logging client before we can set the global logger.
+    global _LOGGER  # pylint: disable=global-statement
```
dparrish: The global statement warning is there for a reason. You should define _LOGGER outside this function (and remove the `global` statement). In fact since this is in all caps this is considered a constant and it should be defined and set at the global scope and never changed.

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -33,48 +32,71 @@
 from gcs_bucket_mover import sts_job_status
 
 _CHECKMARK = u'\u2713'.encode('utf8')
+_LOGGER = None
 
 
-def move_bucket(conf):
+def move_bucket(conf):  # Majority statements are logging. pylint: disable=too-many-statements
```
dparrish: too-many-statements is an indication that your function has too much in it and should be split up into smaller functions. It's VERY difficult to test large functions like this, and you really should have unit tests.

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -85,40 +107,101 @@ def move_bucket(conf):
...
+    _LOGGER.log_text(msg)
+
+    # Turn off any bucket ACLs
+    bucket.acl.save_predefined('private')
+
+    # Revoke all IAM access and then add the service account as an admin
+    account = 'serviceAccount:' + service_account_email
+    policy = bucket.get_iam_policy()
+    for role in policy.keys():
+        for member in policy[role].copy():
```
dparrish: Could you not just do:

```
for role in policy.keys():
  policy[role] = []
```

I looked at the setIamPolicy call and couldn't work out for sure whether it replaces the policy or just updates any roles that you define here. If it just replaces the policy then this whole block isn't needed, you can just declare a new policy and write it.

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -85,40 +107,101 @@ def move_bucket(conf):
...
+            policy[role].discard(member)
+
+    policy['roles/storage.admin'].add(account)
+    bucket.set_iam_policy(policy)
+
+
+def _does_lock_file_exist(bucket, lock_file_name):
```
dparrish: I don't think this function should exist. It doesn't add any value to the caller at all, and the code would be shorter if you just called `storage.Blob(lock_file_name, bucket).exists()` instead of calling this function.

Tho77:
@eshen1991 @dparrish Now that the CLA is signed, I've resubmitted this PR as https://github.com/GoogleCloudPlatform/professional-services/pull/151 and will address David's comments there and this PR can be closed. My apologies for adding this extra step.

ryanmcdowell:
Merged in #151

PR #140, opened 2019-01-07
Tool to detect recency of terraform remote module tags

jmound:
- Git tag recency sorting is handled by `sort`'s `-V`:
https://github.com/GoogleCloudPlatform/professional-services/blob/56425260ba2482e5eb50160636bf9d58c0decd76/tools/terraform-module-update-scanner/check-terraform-module-versions#L46
which is not entirely reliable- it of course depends on the remote authors' versioning scheme. Suggestions are welcome as to how to improve the method of determining the latest tag.


morgante:
Bump.

jmound:
@morgante updated.

morgante:
@jmound Please ping me once you've fixed the test cases.

jmound:
@morgante I've updated the code to handle prefix-styled tags, and added small updates to the tests as well.

PR #151, opened 2019-01-15
Configurable lock, logging to Stackdriver and packaging with Pex.

Tho77:
**Configurable Lock** 

**Stackdriver** 

**Packaging with Pex** 
For ease of use, a Tox ini file was added to enable packaging the tool into a single Pex file that can be run without needing to download any requirements or set up any virtual environments.In order to be able to make a .pex package from MacOS that will run on Linux, a special PyYAML egg is included that is compiled without the OS specific libyaml package.The resulting .pex package is in the /package folder. To support packaging, all environment variables/global testing variables were refactored to be specified in a config yaml file or the command line.


Tho77:
Now that the CLA is signed, I can submit this PR directly.
@dparrish I will address your comments from https://github.com/GoogleCloudPlatform/professional-services/pull/135 in this PR tomorrow.

Tho77:
@dparrish I've updated this PR with your comments from the previous one.

https://github.com/GoogleCloudPlatform/professional-services/pull/135#discussion_r247368685
Unfortunately it's not possible to set the GCP logger client at the global level because the config file with the access key needs to be loaded first. I've removed it from the global scope, set it once in the cli.py and then pass it all over everywhere that needs logging.

https://github.com/GoogleCloudPlatform/professional-services/pull/135#discussion_r247368719
I've broken all of the steps up into sub functions.

https://github.com/GoogleCloudPlatform/professional-services/pull/135#discussion_r247368953
Removed the loop and now it just creates an empty Policy object to assign to the bucket.

https://github.com/GoogleCloudPlatform/professional-services/pull/135#discussion_r247369024
Updated with your suggestions.

PR #167, opened 2019-01-31
Add F5 Rancid Config Parser

jeffmccune:
This patch adds a command line tool to parse F5 configuration data
obtained from Rancid.  The purpose is to transform the F5 configuration,
which may contain many thousands of pools, virtual servers, and iRule
objects, into CSV output which may be easily analyzed using Google
Sheets.

Some select commits from the previous repository:

Add example to list currently active iRules.  Swati requested this
functionality working with the customer.

The examples also include the vip and destination in the outputs.

Add f5 print-irule subcommand

This patch adds a subcommand to print out an irule given a name.  The
purpose is to print out all iRules in a format easily processed.  The
irule can optionally be written to a file and also optionally in
markdown using `--format markdown`

Add jq example to group rules by target pool

This patch adds a jq example grouping all of the results by the target
pool.  The intent is to make it useful to focus on the rules applying to
an application service.

Add map subcommand to output a list of virtual servers

This is a JSON list of objects which can be filtered as per the [Filter
objects based on the contents of a
key](https://github.com/stedolan/jq/wiki/Cookbook#filter-objects-based-on-the-contents-of-a-key)
example in the jq Cookbook.

Add virtuals subcommand to output all virtual server names

This patch adds functionality to print out all virtual server names.
This builds toward mapping target pools to applicable iRules.

Add irules and pools subcommands

This patch adds an irules and pools subcommand to print all iRules given the
F5 configuration.  This establishes the basic structure for parsing the
F5 configuration and extracting information.

Client connections enter a Virtual Server.  iRules are bound to a
virtual server.  Connections are forwarded on to target pools.


jeffmccune:
I signed it!

jeffmccune:
I signed it!

PR #177, opened 2019-02-11
Kunskap

kardiff18:
Added license header to config & changed config variables.


tools/kunskap/tests/main_test.py
```diff
@@ -39,51 +38,39 @@ def setUp(self):
...
         self.mocked_bq().query().result().called
 
-    def testExecuteTransformationQueryWithNull(self):
-        """Tests that query will not execute if there are no usage dates."""
-        main.execute_transformation_query([], self.mocked_bq())
-        self.mocked_bq().query.assert_not_called()
-
     def testPartitionsAndUsageDates(self):
         """Tests that the # of partitions is equal to the # of usage_start_times."""
         bq_client = bigquery.Client()
```
jaketf: This is an integration test that requires table to exist and for the project id etc to be changed in the config. Also it runs queries (sneakily costing the user money). Let's remove this or put in separate file and explain in README.

tools/kunskap/tests/main_test.py
```diff
@@ -39,51 +38,39 @@ def setUp(self):
...
-        main.execute_transformation_query(mocked_dates, self.mocked_bq())
+        main.execute_transformation_query(self.mocked_bq())
         self.mocked_bq().query().result().called
 
-    def testExecuteTransformationQueryWithNull(self):
-        """Tests that query will not execute if there are no usage dates."""
-        main.execute_transformation_query([], self.mocked_bq())
-        self.mocked_bq().query.assert_not_called()
-
     def testPartitionsAndUsageDates(self):
```
jaketf: This is an integration test. Remove or move to separate file and explain the necessary changes to config and required bq environment for this test to pass (table needs to exists etc.)

kardiff18: this test was removed

kardiff18:
conflicting test was removed

PR #189, opened 2019-02-19
Added Tracking Crypocurrencies Exchange Trades with Google Cloud Platform in Real-Time

galic1987:
In this solution, we will discuss the most important design and technical decisions: i) how to set up and configure pipeline for ingesting real-time, time-series data from various crypto exchanges ii) how to design suitable data model, which facilitates querying and graphing at scale. Finally, we will provide a tutorial on how to set up and deploy the proposed architecture using GCP. By following the tutorial steps will establish a connection to multiple exchanges, subscribe to their trade feed, extract and transform these trades into a flexible format to be stored in Bigtable and to be graphed and analyzed.


jaketf:
@galic1987 Can we close this? Looks like this is replaced by #191 .

galic1987:
Correct! @jaketf let's ignore / delete this one

PR #198, opened 2019-02-27
BQ Table Resizer Fix for target_gb Method

annarudy:
- Added float type to --target_gb arg so it can be used in mathematical operation
- Took out reference to self.target_rows before assignment
- Added unit test for resizing using target_tb


annarudy:
@jaketf thanks for catching that! my pycharm had the line size max set to 120. Is there a specific pylint config we use, or just the default?

jaketf:
@annarudy [go/gpylint](https://g3doc.corp.google.com/devtools/gpylint/g3doc/index.md?cl=head)

PR #200, opened 2019-03-05
GCS Bucket Mover: Rename Feature

Tho77:
- Added a new flag --target_bucket_name that allows the user to rename a GCS bucket instead of or in addition to a project move.
- Changed the bucket lock option to be enabled by default and updated the default lock file name.
- Refactored the command line options to adhere to a common naming style.


tools/gcs-bucket-mover/config.yaml
```diff
@@ -1,7 +1,7 @@
 gcp_source_project_service_account_key: </PATH/TO/YOUR/SOURCE-PROJECT_SERVICE_ACCOUNT_KEY.JSON>
 gcp_target_project_service_account_key: </PATH/TO/YOUR/TARGET-PROJECT_SERVICE_ACCOUNT_KEY.JSON>
 
-lock_file_name: .locked
+lock_file_name: _locks/all.lock
```
blueandgold: How is the `_locks/` created?  Will there be any permission issues?

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -62,32 +55,68 @@ def move_bucket(config, parsed_args, cloud_logger):
...
-    _assign_sts_permissions_to_new_bucket(cloud_logger, sts_account_email,
-                                          config)
-    _run_and_wait_for_sts_job(sts_client, config.target_project,
-                              config.temp_bucket_name, config.bucket_name,
-                              cloud_logger)
-
-    _delete_empty_temp_bucket(cloud_logger, target_temp_bucket)
-    _remove_sts_permissions(cloud_logger, sts_account_email, config)
+    if config.is_rename:
+        target_bucket = _create_target_bucket(cloud_logger, config,
```
blueandgold: As discussed offline, it would help with maintainability and readability to decompose this into separate `move_bucket()` and `rename_bucket()` helper methods, and change the current method to `main()`.

Tho77: Refactored into smaller, more logical methods.

tools/gcs-bucket-mover/config.yaml
Tho77: The locks file would be created by some other process, and that process would need to ensure the object is visible to the tool.
Since the tool requires permission to list the source bucket, unless there is special object level permissions, there shouldn't be any issues.

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -856,3 +923,7 @@ def _write_spinner_and_log(spinner, cloud_logger, message):
     """
     spinner.write(message)
     cloud_logger.log_text(message)
+
+
+if __name__ == '__main__':
+    main()
```
blueandgold: Add a newline.

tools/gcs-bucket-mover/gcs_bucket_mover/bucket_mover_service.py
```diff
@@ -140,29 +200,34 @@ def _lock_down_bucket(spinner, cloud_logger, bucket, lock_file_name,
     bucket.set_iam_policy(policy)
 
 
-def _create_temp_target_bucket(cloud_logger, config, source_bucket_details):
-    """Create the temp bucket in the target project
+def _create_target_bucket(cloud_logger, config, source_bucket_details,
+                          bucket_name):
+    """Creates the temp bucket (or target bucket during rename) in the target project
```
blueandgold: Is the `temp` in the doc string still needed?

PR #204, opened 2019-03-09
Updates to the notebook so it can be used in ML notebooks

smhosein:
This PR mainly contains minor updates so that it can be used seamlessly in ML Notebooks as well. There are a few spelling mistakes taken care of also.


hashkanna:
Fix the GCS BUCKET locations to point to dummy ones

PR #207, opened 2019-03-12
Initial add of CloudSQL custom metric

kevensen:
Example of how to create a custom metric for Stackdriver Monitoring that counts CloudSQL private instances.


examples/cloudsql-custom-metric/main.py
```diff
@@ -0,0 +1,352 @@
...
+    series.resource.labels['namespace'] = row.privatenetwork
+    point = series.points.add()
+    if row.availabilitytype == "ZONAL":
+        point.value.int64_value = 1
+    elif row.availabilitytype == "REGIONAL":
+        point.value.int64_value = 1
+
+    now = time.time()
+    point.interval.end_time.seconds = int(now)
+    # point.interval.end_time.nanos = int(
```
jaketf: remove commented out code.

examples/cloudsql-custom-metric/requirements.txt
```diff
@@ -0,0 +1,5 @@
+google-cloud-bigquery==1.9.0
```
jaketf: Unpin dependencies if explicit versions not needed.

examples/cloudsql-custom-metric/main.py
```diff
@@ -0,0 +1,352 @@
...
+        cloud_logger.debug("running forever")
+        timeseries = get_cloudsql_timeseries(
+                                    project_id,
+                                    dataset,
+                                    True,
+                                    location)
+        create_timeseries(project_id, timeseries)
+
+        ticker = threading.Event()
+        while not ticker.wait(WAIT_TIME_SECONDS):
```
jaketf: Is there a reason to have a long running thread to do this forever? It seems there's an opportunity to use a [Cron Job](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/). This would prevent one failure from killing subsequent runs and make it clearer what failed when.

examples/cloudsql-custom-metric/main.py
```diff
@@ -0,0 +1,352 @@
...
+                timeseries = get_cloudsql_timeseries(
+                                    project_id,
+                                    dataset,
+                                    False,
+                                    location)
+            create_timeseries(project_id, timeseries)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Query BigQuery for \
```
jaketf: Convention nitpicking: Let's move the parser logic to main method. Refactor this to be: 
```
if __name__ == "__main__":
   main()
```
and then main should take no arguments and do this argument parsing there.

examples/cloudsql-custom-metric/main.py
kevensen: The challenge I have here is that on the first run, I want the query to be different.  I'll think about this.

examples/cloudsql-custom-metric/requirements.txt
```diff
@@ -1,5 +1,4 @@
-google-cloud-bigquery==1.9.0
-google-cloud-monitoring==0.31.1
 google-cloud-logging
```
jaketf: please specify minimum versions of these libraries.

kevensen:
Hi @jaketf.  I appreciate the review, thank you.

jaketf:
@kevensen can you update the top level README with a pointer to this example?

kevensen:
@jaketf I've added the link

jaketf:
LGTM. Nice work!

PR #211, opened 2019-03-14
BigQuery Org-Wide Job Complete Notifier

glickbot:
Terraform scripts and code to setup:
Org-wide aggregated log export -> PubSub for BQ Job Complete events
Cloud Function to listen -> extract metadata -> possibly notify based on event
Source Repository as source to function, with a Cloud Builder trigger to redeploy the function
Intended to be run/managed centrally for a large organization, allowing methods to use job metadata to notify the right teams/project, etc


tools/bq-notifier/main.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// HelloPubSub consumes a Pub/Sub message.
+func HandleJobComplete(ctx context.Context, m PubSubMessage) error {
+	var data interface{}
+	if err := json.Unmarshal(m.Data, &data); err != nil {
+		log.Fatalf("Unable to unmarshal Job Information: %v", err)
+	}
+	var job Job
+	err := mapstructure.Decode(data, &job)
```
sabhyankar: if err := mapstructure.Decode(data, &job); err != nil {
    return err;
}

tools/bq-notifier/main.go
```diff
@@ -0,0 +1,69 @@
...
+	bqn.AddHook(SlackNotify)
+}
+
+type PubSubMessage struct {
+	Data []byte `json:"data"`
+}
+
+// HelloPubSub consumes a Pub/Sub message.
+func HandleJobComplete(ctx context.Context, m PubSubMessage) error {
+	var data interface{}
```
sabhyankar: Maybe just get a pointer to an interface since we are using it by reference in two of the calls below?

tools/bq-notifier/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+package bqnotifier
+
+import (
+	"context"
+	"io/ioutil"
+	"log"
+	"testing"
+)
+
+func TestJson(*testing.T) {
```
sabhyankar: We want to return test errors using the testing.T reference that is available. So we would do something like this..

func TestJson(t *testing.T) {
   ... setup and run the test ...
  if err != nil {
    t.Errorf("'%s' failed: HandleJobComplete(%v, %v) = (%v) want (nil)", name, input1, input2, err)
  }
}

tools/bq-notifier/slack.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	successColor = "#7CD197"
+	errorColor = "#F35A00"
+	slackFmt = "https://hooks.slack.com/services/%s/%s/%s"
+	tokenSplit = "-"
+
+)
+
+// Please note: this is an example and hasn't been fully tested.
```
sabhyankar: Add a description in addition to the note.

tools/bq-notifier/slack.go
```diff
@@ -0,0 +1,74 @@
...
+func SlackNotify(job Job, _ interface{}) error {
+	labels := job.Labels()
+	if token, ok := labels["bqnlabel"]; ok {
+		tokens := strings.Split(token, tokenSplit)
+		if len(tokens) != 3 {
+			return errors.New(fmt.Sprintf("Invalid token format, recieved %s", token))
```
sabhyankar: return fmt.Errorf(..)

tools/bq-notifier/slack.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		var color string
+		if job.IsError() {
+			color = errorColor
+		} else {
+			color = successColor
+		}
+		jsonStr := []byte(fmt.Sprintf(`
```
sabhyankar: Might be cleaner to just create a local struct and use json.Marshal

tools/bq-notifier/slack.go
```diff
@@ -0,0 +1,74 @@
...
+			return err
+		}
+		req.Header.Set("Content-Type","application/json")
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != 200 {
+			b, _ := ioutil.ReadAll(resp.Body)
+			return errors.New(fmt.Sprintf(
```
sabhyankar: return fmt.Errorf(..)

tools/bq-notifier/slack.go
```diff
@@ -0,0 +1,74 @@
...
+				"Error posting to slack url, non 200 code of %d found posting to url: %s, response: %s",
```
sabhyankar: Generally we should avoid starting the error string with an upper case: See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

tools/bq-notifier/README.md
```diff
@@ -0,0 +1,77 @@
...
+* Setup application-default auth for terraform* (unnecessary on cloud shell)
+* [Install Golang](https://golang.org/doc/install)
+* Setup git config
+    * git config --global user.email "you@example.com"
+    * git config --global user.name "Your Name"
+\* Using application-default auth for terraform isn't the best practice, but it's the easiest to setup. For Production terraform use, please follow terraform on GCP best practices.
+
+## Setup
+* Aquire code (git clone this repository and move to the directory this README is in.
+* Copy ```terraform.tfvars-example``` to ```terraform.tfvars```, and modify it's contents.
```
jasonmar: Only need single backticks these

tools/bq-notifier/install.tf
```diff
@@ -0,0 +1,140 @@
...
+    "${google_logging_organization_sink.jobcomplete_sink.writer_identity}",
+  ]
+}
+
+resource "google_project_service" "source_repo_api" {
+  project = "${var.project}"
+  service = "sourcerepo.googleapis.com"
+}
+
+resource "google_sourcerepo_repository" "function_repo" {
```
jasonmar: This one runs a script and it's not immediately clear to me what the script is doing, could you add a comment explaining its purpose?

tools/bq-notifier/install.tf
```diff
@@ -0,0 +1,140 @@
+# This code is a prototype and not engineered for production use.
+# Error handling is incomplete or inappropriate for usage beyond
+# a development sample.
+
+variable "topic" {
```
jasonmar: Moving the variables declarations to a separate file would make this easier to read

tools/bq-notifier/configure_repo.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# This code is a prototype and not engineered for production use.
```
jasonmar: Can you add a comment on the purpose of the script here?

tools/bq-notifier/README.md
```diff
@@ -0,0 +1,77 @@
...
+## Overview
+
+The terraform scripts included will activate and use the following APIs:
+* Cloud Functions
+* Source Repositories
+* Cloud Builder
+   
+The overall architecture is:
+* Org-wide aggregated log sync of BigQuery JobComplete events with a pubsub topic destination
+* A cloud function in Golang that triggers off the pubsub topic, who's source is a branch of a cloud repository
```
jasonmar: It would be helpful to expand on how you're expecting branches to be used.

PR #219, opened 2019-03-22
mwallman gce to adminsdk

mwallman:
Example code for accessing Google AdminSDK from GCE service account identity, with basic instruction on delegated subject setup.


bmenasha:
Looks great. thanks for creating. I submitted two small changes.

PR #222, opened 2019-03-27
Documentation Update

bmenasha:
Update readme documentation  (although I'm a bit confused why commits 9ee71ee and 1a073a2 are claimed as being merged, they should have been merged before.


tools/asset-inventory/gae/config.yaml
```diff
@@ -13,19 +13,55 @@
...
+import_dataflow_project: <ENTER-PROJECT>
+
+# When using the template runner. Supply values as a json representation of
+# RuntimeEnvironment:
+# https://cloud.google.com/dataflow/docs/reference/rest/v1b3/RuntimeEnvironment
+# If you are running the Dataflow workers in a shared VPC, set the VPC network
+# and subnet here.
+import_pipeline_runtime_environment: >
+  {
+    # if you don't get a warning about firewall rules preventing shuffle
```
jaketf: These comments should be moved to the troubleshooting section of the README, not burried in a config file. I would link out to the [docs on dataflow routes and firewall](https://cloud.google.com/dataflow/docs/guides/routes-firewall). I typically recommend they make a dataflow subnet with these rules and specify this when running their pipeline. Makes it easier to run with private ip as well.

bmenasha: great. thanks for looking. 
I made the changes.

PR #230, opened 2019-04-05
Fixing PR 185

jaketf:

PR #234, opened 2019-04-09
Add RECORD type support for parquet.

jaketf:
This PR integrate the work from #230 and #202 to provide support for generating nested types even when using the parquet output format.

### DIFF
--- a/examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
+++ b/examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
@@ -1,13 +1,13 @@
 # Copyright 2018 Google Inc.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0 (the 'License');
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
 #      http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
+# distributed under the License is distributed on an 'AS IS' BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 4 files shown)


jaketf:
@annarudy Can you review this?

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
```diff
@@ -25,30 +25,56 @@ def get_pyarrow_translated_schema(string_schema):
...
+            'INTEGER': pa.int64(),
+            'FLOAT': pa.float64(),
+            'NUMERIC': pa.int64(),
+            'BOOLEAN': pa.bool_(),
+            'TIMESTAMP': pa.timestamp('us'),
+            'DATE': pa.date32(),
+            'TIME': pa.time64('us'),
+            'DATETIME': pa.timestamp('us'),
+            'GEOGRAPHY': None,
+            'RECORD': None
```
annarudy: We can take the 'RECORD': None out now. I threw it in there so the ValueError would be raised below if someone tried to use a RECORD type. but now that you've added the support, we shouldn't need it!

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
```diff
@@ -25,30 +25,56 @@ def get_pyarrow_translated_schema(string_schema):
...
+        :param field (bigquery.schema.SchemaField)
+        :return: pa.DataType
+        """
+        type_conversions = {
+            'STRING': pa.string(),
+            'NUMERIC': pa.int64(),
+            'BYTE': None,
+            'INTEGER': pa.int64(),
+            'FLOAT': pa.float64(),
+            'NUMERIC': pa.int64(),
```
annarudy: This is totally my fault, but I listed NUMERIC twice. Would you mind deleting this one?

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
```diff
@@ -25,30 +25,56 @@ def get_pyarrow_translated_schema(string_schema):
...
+                return pa.list_(
+                    _bq_to_pa_type(type_conversions.get(field['type'])))
+        elif field['type'] == 'RECORD':
+            nested_fields = field['fields']
+            # Recursively call to convert the next nested layer.
+            return pa.struct(
+                [(fld['name'], _bq_to_pa_type(fld))
+                    for fld in nested_fields])
+        else:
+            return type_conversions.get(field.get('type'))
```
annarudy: Just curious why you used the get() method on the dict here rather than dict[key] method (type_conversions[field['type']]) since the dict[key] method is used elsewhere

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
jaketf: removed.

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
jaketf: removed.

examples/dataflow-data-generator/data-generator-pipeline/data_generator/ParquetUtil.py
jaketf: Good catch. Using `dict.get` returns `None` if the key doesn't exist while `dict[key]` will throw a `KeyError`. If it gets here and the type is not present in `type_conversions I should raise an exception.

annarudy:
It looks like I'll need write access to be able to properly approve the PR. I'll look into that.

PR #245, opened 2019-05-02
Add Cloud Function that deletes VM instances created without CMEK

zachberger:
This Cloud Function detects newly created VM Instances that have disks attached
with a Google-managed encryption key attached and then immediately deletes
any such VM instances.


zachberger:
Replaced by terraform-google-modules/terraform-google-event-function#17

PR #270, opened 2019-06-20
Add GCF VM Delete Event Handler Example

jeffmccune:
Customer use case: When a VM is deleted, DNS A records associated with the VM should be automatically deleted so that stale DNS records do not pile up.

The customer requested the solution be general enough to support VMs deleted through the Console UI, gcloud, Deployment Manager, Terraform, etc...

The solution uses a GCF Background Function handling `compute.instances.delete` events.  The customer also required the function compare the IP address of the VM's primary interface with the DNS A record rrdatas to guard against deleting the incorrect record.  This IP check introduces a race condition, the function races against the VM delete operation.  In practice the function wins the race most of the time and the cleanup is successful.


jeffmccune:
b/135707396

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+import os
+import google.cloud.logging
+from enum import Enum
+from google.cloud.logging.resource import Resource
+from googleapiclient import discovery
+from googleapiclient.errors import HttpError
+from typing import List
+
+
+class RuntimeState:
```
jaketf: Confused to the value of this empty class.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+
+    A NOT_PROCESSED results is likely a result of losing the race against the
+    VM delete operation and is intended to signal to the user they may need to
+    cleanup DNS records using another mechanism (e.g. manually).
+    """
+    OK = 0
+    NOT_PROCESSED = 1
+
+
+class Severity(Enum):
```
jaketf: This file is rather long, can we move all these logging artifacts [into a package](https://cloud.google.com/functions/docs/writing/#functions-writing-file-structuring-python)? 
- StructuredLog and subclasses
- These Enum subclasses.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+    def load_configuration(self):
+        """Loads configuration from the environment
+
+        Returns:
+          Dictionary of config key/values.
+        """
+        dns_project = os.getenv('DNS_VM_GC_DNS_PROJECT')
+        if not dns_project:
+            raise(EnvironmentError(
+                'Env var DNS_VM_GC_DNS_PROJECT is required.'
```
jaketf: Use full words in Error messages.
```python
raise(EnvironmentError(
    'Environment Variable DNS_VM_GC_DNS_PROJECT is required.'
))
```

Also can we refactor this env var to not have `DNS` twice
s/DNS_VM_GC_DNS_PROJECT/DNS_VM_GC_PROJECT/

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+            ))
+        dns_zones = os.getenv('DNS_VM_GC_DNS_ZONES')
+        if not dns_zones:
+            raise(EnvironmentError('Env var DNS_VM_GC_DNS_ZONES is required'))
```
jaketf: (same as comment above)

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+                metadata.
+        Returns:
+            Number of records deleted across all managed zones.
+        """
+        handler = EventHandler(app=self, data=data, context=context)
+        result = handler.run()
+        return result
+
+
+def main(data, context=None, http=None, session=None):
```
jaketf: Wrapping this in a `main` method seems to be an unnecessary level of abstraction and IMO makes your tests less readable with `main.main`.

Instead you should bump this RuntimeState logic directly into the `dns_vm_gc` and test that method.
I also think `main.py` should have a more descriptive name. You could change this to `dns_vm_gc.py`

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+
+
+class StructuredLog(object):
+    """Base class to report structured log events
+
+    Sub classes are expected to override MESSAGE (str), RESULT (Result), and
+    DETAIL (Detail).
+
+    """
+    MESSAGE = '{} mark'
```
jaketf: what is the  `mark` for?

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+        self.function_name = os.getenv('FUNCTION_NAME')
+        stream = os.getenv('DNS_VM_GC_REPORTING_LOG_STREAM')
+        # The log stream structured reports are sent to.
+        # https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
+        if stream:
+            self.log_name = stream
+        else:
+            self.log_name = (
+                'projects/{}/logs/{}'
+            ).format(self.function_project, self.function_name)
```
jaketf: Nit: use f string for brevity / readability.
GCF run python 3.7
As of python 3.6 we have formatted string literals [PEP 498](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498)
```python
self.log_name = f"projects/{self.function_project}/{self.function_name}"
```

Cosider using this instead of `str.format()` throughout this file.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+            'message': self.message(),
+            'vm_uri': self.vm_uri,
+            'result': self.result(),
+            'detail': self.detail(),
+        }
+
+    def log_entry(self):
+        """Assembles dict intended for use as LogEntry
+
+        This structure is generally passed as keyword arguments to the
```
jaketf: remove "generally" it's confusing.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+        See: https://cloud.google.com/dns/docs/reference/v1/resourceRecordSets
+        """
+        super().__init__(vm_uri, event_id)
+        self.dns_project = dns_project
+        self.dns_managed_zone = dns_managed_zone
+        self.record = record
+        self.response = response
+
+    def message(self):
+        return self.MESSAGE.format(self.vm_uri, self.record['name'])
```
jaketf: If we're overriding this `message` method,  might as well define Message in line here with an f string:
```suggestion
        return f"{self.vm_uri} matches DNS record {self.record['name']} deleted (RR_DELETED)"
```

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+class NameMismatch(RecordSkipped):
+    DETAIL = Detail.RR_NAME_MISMATCH
+
+
+class IpMismatch(RecordSkipped):
+    DETAIL = Detail.RR_IP_MISMATCH
+
+
+class RecordDeleted(StructuredLog):
+    MESSAGE = "{} matches DNS record {} deleted (RR_DELETED)"
```
jaketf: remove, see comment below. 
You're overriding `message` method so defining this class attribute and overriding the only method is just code sprawl.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+class VmNoIp(StructuredLog):
+    MESSAGE = "{} has no IP address (VM_NO_IP)"
+    SEVERITY = Severity.INFO
+    RESULT = Result.OK
+    LEVEL = logging.INFO
+    DETAIL = Detail.VM_NO_IP
+
+
+class RecordSkipped(StructuredLog):
+    MESSAGE = "{} does not match DNS record {} ({})"
```
jaketf: see comments on `RecordDeleted` same apply here.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+            managedZone=managed_zone,
+            body=change)
+        response = request.execute()
+
+        event = RecordDeleted(self.vm_uri, self.event_id, project,
+                              managed_zone, record, response,)
+        self.log_event(event)
+        return response
+
+    def find_garbage(self,
```
jaketf: more descriptive name 
s/find_garbage/find_garbage_dns_entries/

examples/gcf-pubsub-vm-delete-event-handler/requirements.txt
```diff
@@ -0,0 +1,3 @@
+google-api-python-client==1.7.9
```
jaketf: do we need to pin all these dependencies?

jeffmccune: No need, just habit.  Will remove the version spec entirely, please comment if there's a more preferred way (e.g. specifying to major or minor version, but not bugfix).

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: It's intended to hold the App instance at runtime and provide a convenient way to test the behavior of warm/cold starts from pytest via modifying `RuntimeState.app`.

Happy to change it if there's a more preferable way to accomplish the same, but a class attribute seemed to be the simplest way to achieve this.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Yes.  Related, the customer reported an issue.  Their managed zones are spread across multiple VPC networks, so I'd like to remove this configuration setting and instead make `DNS_VM_GC_DNS_ZONES` a list of fully qualified URI's.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: We could, but DNS was repeated intentionally.  All configuration settings are prefixed with DNS_VM_GC_ for operational convenience, and I kept getting confused about "zones" referring to either the DNS managed zone or the GCE compute zone, which is why I arrived at DNS_VM_GC_DNS_ZONES to indicate the managed zones.

Will change this to `DNS_VM_GC_MANAGED_ZONES` which removes the duplication and resolves the ambiguity with compute zones.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: It comes from the convention of logging nothing having the behavior of a "mark" syslog entry marking the fact that nothing was logged.

Sub classes should always override this attribute, so I'll remove it entirely.  With the other changes, there will be an error when `self.MESSAGE.format` is called in the sub-class if it's not overridden.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Ack, will do.  Thanks for letting me know, it's been ~10 years since I last wrote serious Python.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Ack, will refactor all of the StructuredLog classes to use inline message format strings.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Sure thing

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+            result = self.compute.instances().get(
+                project=project,
+                zone=compute_zone,
+                instance=instance).execute()
+        except HttpError as err:
+            self.log.error("Getting {}: {}".format(self.vm_uri, err))
+            result = {}
+        return result
+
+    def validate_data(self, data):
```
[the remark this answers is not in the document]
jeffmccune: Good catch, I was anticipating more validation, but YAGNI prevailed again.

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+        """Stackdriver Log Severity
+
+        https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
+        """
+        return self.SEVERITY.name
+
+    def info(self):
+        """Assembles dict intended for jsonPayload"""
+        return {
+            'message': self.message(),
```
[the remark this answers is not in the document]
jeffmccune: Will take another look at this, but I'm a bit confused on where the behavior of interpolating the message should move to.  The initializer?

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+            'project_id': self.function_project,
+            'region': self.function_region,
+        }
+        resource = Resource(labels=resource_labels, type='cloud_function')
+        log_entry = {
+            'log_name': self.log_name,
+            'labels': {
+                'event_id': self.event_id,
+            },
+            'severity': self.severity(),
```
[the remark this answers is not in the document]
jeffmccune: Ack

examples/gcf-pubsub-vm-delete-event-handler/requirements.txt
jaketf: I typically recommend just specify a minimum version.

examples/gcf-pubsub-vm-delete-event-handler/main.py
jaketf: I think "mark" in stackdriver is just log noise. 

You can specify an abstract attribute like so 
```python
   # In base class
    @property
    @abstractmethod
    def MESSAGE(self):
         raise NotImplementedError
...
    # In subclass
    MESSAGE = 'string for formatting {}'
```

examples/gcf-pubsub-vm-delete-event-handler/main.py
jaketf: Sure thing! New features are hard to keep track of even when you write python every day!

examples/gcf-pubsub-vm-delete-event-handler/main.py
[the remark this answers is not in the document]
jaketf: see updated comment just do the format right here in `info(self)`

examples/gcf-pubsub-vm-delete-event-handler/main.py
jaketf: Nice, that makes a lot of sense.

examples/gcf-pubsub-vm-delete-event-handler/main.py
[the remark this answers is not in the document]
jeffmccune: Understood thanks.  I'm fixing all of these up now.

I'm going to keep message() as the sole getter because some of the structured logs take more than just the vm_uri.  For example:

```python
class RecordSkipped(StructuredLog):
    MESSAGE = "{} does not match DNS record {} ({})"

    def message(self):
        return self.MESSAGE.format(self.vm_uri, self.record['name'], self.DETAIL.name)
```

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+    VM_NO_IP = 3
+    IGNORED_EVENT = 4
+    LOST_RACE = 5
+    RR_MISMATCH = 6
+    RR_NOT_A_RECORD = 7
+    RR_NAME_MISMATCH = 8
+    RR_IP_MISMATCH = 9
+
+
+class StructuredLog(object):
```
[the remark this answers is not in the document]
jeffmccune: Fixed in 4a4326bbd7332ebf3e96d764ef60fbdd8965243a

examples/gcf-pubsub-vm-delete-event-handler/main.py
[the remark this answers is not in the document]
jeffmccune: Fixed in 4a4326bbd7332ebf3e96d764ef60fbdd8965243a

examples/gcf-pubsub-vm-delete-event-handler/main.py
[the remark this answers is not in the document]
jeffmccune: Fixed in 4a4326bbd7332ebf3e96d764ef60fbdd8965243a

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+        Events must match the following filter to trigger the cleanup process:
+
+                resource.type="gce_instance"
+                jsonPayload.event_type="GCE_API_CALL"
+                jsonPayload.event_subtype="compute.instances.delete"
+
+        Returns (bool): True if the event should be handled.
+        """
+        if event_type == 'GCE_API_CALL':
+            if event_subtype == 'compute.instances.delete':
```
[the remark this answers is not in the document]
jeffmccune: Fixed in 4a4326b..5145e75

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+                'event_id': self.event_id,
+            },
+            'severity': 'INFO',
+            'resource': resource,
+        }
+        log_entry.update(kw)
+        jsonPayload['message'] = msg
+        self.cloud_log.log_struct(info=jsonPayload, **log_entry)
+
+    def dns_records(self, project: str, managed_zone: str) -> List[dict]:
```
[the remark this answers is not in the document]
jeffmccune: Fixed in 5145e75..337be82

examples/gcf-pubsub-vm-delete-event-handler/main.py
```diff
@@ -0,0 +1,660 @@
...
+        return candidates
+
+    def ip_address(self, instance):
+        """Parses the primary network IP from a VM instance Resource.
+
+        Args:
+        Returns: (string) ip address or None if IP not found
+        """
+        ip = None
+        if 'networkInterfaces' in instance:
```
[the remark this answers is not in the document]
jeffmccune: Fixed in 337be82..23f3836

examples/gcf-pubsub-vm-delete-event-handler/main.py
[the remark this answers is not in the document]
jeffmccune: Fixed in 23f3836..f668eb6

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed in f668eb6..20d3bb1

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed in 20d3bb1..ecde344

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed in 20d3bb1..ecde344

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed in ecde344...5eef9da

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed in ecde344...5eef9da

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Fixed

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Decided to keep this as is because the new use of ABC requires MESSAGE to be overridden anyway.  In this case, I think it makes sense to override message() as well because there are additional format field for this subclass.

jeffmccune: Ah, this finally sunk in.  Changing this to get ride of MESSAGE entirely and just always override message().

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Done

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Done

examples/gcf-pubsub-vm-delete-event-handler/main.py
jeffmccune: Done

examples/gcf-pubsub-vm-delete-event-handler/requirements.txt
jeffmccune: Done

jeffmccune:
@jaketf Thanks a ton for the in-depth review, I'm grateful for your time and feedback.  I've addressed all the requested changes and tested out the DNS cleanup still works.  Please take another look at your convenience.

PR #273, opened 2019-06-23
Update command on e2e-home-appliance-status-monitoring example

tvdias:
Since step 0 already steps into `examples/e2e-home-appliance-status-monitoring`, on step 2 we only need to step into `server`.


ryanmcdowell:
@tvdias -- Any timeline to make final changes to this PR for it to not be in draft?

PR #277, opened 2019-07-03
Added ThrottlingStrategy and TimeDomain.timer value to DataflowThrottling library

SatishAthukuri:
* ThrottlingStrategy : This defines a strategy that is applied when a request is throttled by server or client.

* TimeDomain.timer value : This sets Time domain value to EVENT_TIME/PROCESSING_TIME using the PCollection type [Bounded/UnBounded].


tools/dataflow-throttling/README.md
```diff
@@ -32,6 +33,9 @@ Pipeline PCollection transform steps:
 
 The variables which stores the total number of requested and accepted requests are zeroed out after a certain interval of time (defaults to 1 min). This reset speeds up client recovery if server throughput was limited due to limitations on the server side not caused by the client.
 
+### Limitations
+* When the Dynamic Throttling Transform is used inside FixedWindows, the state is restricted per key per window and rejection probability as well as the throughput rate is calculated per window. This can cause an overlap of a fixed window that processes the current events with the other fixed windows which are processing backlog; thus causing a higher rate of events sent to the external service.
```
zhaber: When the Dynamic Throttling Transform is used inside FixedWindows, the state is restricted per key per window and rejection probability as well as the throughput rate is calculated per window. This can cause an overlap of a fixed window that processes the current events with other fixed windows which are processing a backlog; thus causing a higher rate of events sent to the external service.

PR #290, opened 2019-07-22
updates to dataflow python examples

ShobhaRaniD:
- update dependency `apache-beam` to the newer version to avoid the "SSLHandshakeError: [SSL: CERTIFICATE_VERIFY_FAILED]" error
- fixing some typos and other refactors for better readability


examples/dataflow-python-examples/requirements.txt
```diff
@@ -1,4 +1,4 @@
-apache-beam[gcp]==2.4.0
+apache-beam[gcp]==2.11.0
```
jaketf: Let's unpin this

examples/dataflow-python-examples/requirements.txt
```diff
@@ -1,4 +1,4 @@
...
 Faker==0.8.13
```
jaketf: These `Faker` and `faker-schema` things are unused. remove.

examples/dataflow-python-examples/requirements.txt
```diff
@@ -1,4 +1,4 @@
...
 faker-schema==0.1.4
 google-cloud==0.32
```
jaketf: Let's unpin / update this as well.

examples/dataflow-python-examples/requirements.txt
ShobhaRaniD: Done.

examples/dataflow-python-examples/requirements.txt
ShobhaRaniD: Done.

examples/dataflow-python-examples/requirements.txt
ShobhaRaniD: Done.

ShobhaRaniD:
> See some minor suggestions on further improvements to `requirements.txt`

👍 

> For extra credit points, could you see if these pipelines are forwards compatible w/ python3 ?

There are quite a few changes that this code requires for porting to and testing with python 3 (especially syntactical differences). I think a good time to update those would be when Apache beam is fully supported on py 3.6, which it seems like should be pretty soon - https://issues.apache.org/jira/browse/BEAM-1251?focusedCommentId=16890504&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16890504. I can get back to this in a future PR.

jaketf:
Ok Makes sense, LGTM. 
I think beam plans on [python3 support in 2.14.0](https://github.com/apache/airflow/pull/5602#issuecomment-515874853)

PR #296, opened 2019-08-09
Fix issue with incorrect image filename generated in annotations

cfezequiel:
This fixes a bug in `annotation.read_from_pascal` where the image filename is ignored, resulting one of the callers, `dataset.gen_csv_from_annotations` using the annotation filename instead in the output CSV. 

This also adds `os.path.expanduser` calls whenever a file or directory is read to support home-relative directory paths.

Eng. Council bug ID: b/137942099

### DIFF
--- a/tools/cloud-vision-utils/cloud_vision_utils/annotation.py
+++ b/tools/cloud-vision-utils/cloud_vision_utils/annotation.py
@@ -83,30 +83,35 @@ def read_from_pascal(filename: str):
   Args:
     filename: PASCAL VOC XML filename.
 
-  Yields:
-    `BoundingBox` objects.
+  Returns:
+    (image filename, list of `BoundingBox` objects).
   """
 
   with gfile.GFile(filename) as f:
     tree = ET.parse(f)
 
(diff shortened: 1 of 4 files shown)


tools/cloud-vision-utils/cloud_vision_utils/annotation.py
```diff
@@ -83,30 +83,35 @@ def read_from_pascal(filename: str):
...
-    `BoundingBox` objects.
+  Returns:
+    (image filename, list of `BoundingBox` objects).
   """
 
   with gfile.GFile(filename) as f:
     tree = ET.parse(f)
 
   root = tree.getroot()
+  image_filename = root.find('filename').text
```
EricBeach: can we be guaranteed this will be found?

cfezequiel: Image labeling tools that support the PASCAL VOC format should include the source image 'filename' in the output annotation XML file. If the 'filename' element is not found then this function will throw an exception.

PR #298, opened 2019-08-14
Add content-based features to model and change to classification

TheMichaelHu:
- add a genre vector feature to the users' features
- change the optimization problem from regression to classification, where the label is now whether or not a user has listened to a song more than twice
- change evaluation metric from recall of top 10 to precision of each user-item pair from a random sample of 1000 items.

### DIFF
--- a/examples/cloudml-collaborative-filtering/preprocessing/preprocess.py
+++ b/examples/cloudml-collaborative-filtering/preprocessing/preprocess.py
@@ -44,6 +44,17 @@ def ReadBQ(p, query):
       query=query, use_standard_sql=True))
 
 
+def _normalize_user_tags(features):
+  """Fills in missing user tags features and normalizes the tag vectors."""
+  tags_sum = sum(features[constants.USER_TAGS_KEY])
+  if tags_sum > 0:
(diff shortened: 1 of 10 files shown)


examples/cloudml-collaborative-filtering/preprocessing/preprocess.py
```diff
@@ -44,6 +44,17 @@ def ReadBQ(p, query):
       query=query, use_standard_sql=True))
 
 
+def _normalize_user_tags(features):
```
jaketf: nit: functions should do one thing. split this in two and add unittests please.

examples/cloudml-collaborative-filtering/preprocessing/preprocess.py
```diff
@@ -44,6 +44,17 @@ def ReadBQ(p, query):
...
+  """Fills in missing user tags features and normalizes the tag vectors."""
```
jaketf: This docstring should describe the type/structure of the expected input.

nit: "fills in" implies that you are mutating the input value in-place but this function actually returns the modified value.

PR #308, opened 2019-08-30
Prepare package for PyPI publishing

ocervell:
This PR adds the `long_description_content_type` need for a valid upload to PyPI


jaketf:
@ocervell if you are actually hosting this in PyPi It should be migrated into it's own repo.

PR #309, opened 2019-09-02
auto detect source and fix transform bug

luotigerlsx:
This PR fix the following two issues:
1. auto detection source for bigquery table or text file
2. inconsistency in transform_utils


examples/cloudml-sklearn-pipeline/trainer/task.py
```diff
@@ -75,8 +75,14 @@ def run_experiment(flags):
   """Testbed for running model training and evaluation."""
   # Get data for training and evaluation
 
-  dataset = utils.read_df_from_bigquery(
+  # If BigQuery table, specify as as PROJECT_ID.DATASET.TABLE_NAME
+  if len(flags.input.split('.')) == 3:
```
jaketf: You could consider handling the legacy sql table reference `project:dataset.table` which is what the `bq` cli still accepts.

luotigerlsx: Hi @jaketf , thanks a lot for approving. I actually make the standard sql as the requirement while initiating the client.

PR #313, opened 2019-09-11
Fixing syntax to add open parentheses

mservidio:
Adds in missing parentheses


mservidio:
@freedomofnet Could you please review?

PR #314, opened 2019-09-11
Oracle to BigQuery replication pipeline

zhaber:


ryanmcdowell:
Re-opening as I think we should still pursue this.

PR #315, opened 2019-09-12
Adding CI testing for tools - linting python files with yapf

AdrienWalkowiak:
- Adding a "test" target for the Makefile - this errors out if any file not under the folder exclusion list needs linting (according to yapf using google style).
- All current folders under tools are excluded from this test at this point (only new ones will trigger)
- Adding Dockerfile and Cloudbuild conf file to integrate


cloudbuild/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# This Dockerfile builds the image used in Cloud Build CI to run 'make test'.
+# To push a new image, run 'gcloud builds submit --project=cloud-eng-council --tag gcr.io/cloud-eng-council/make .'
+# from this directory.
+
+FROM debian
+
+RUN apt-get update && apt-get install -y build-essential python python-pip
+RUN pip install yapf
```
morgante: Should we use Python 3?

cloudbuild/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# This Dockerfile builds the image used in Cloud Build CI to run 'make test'.
+# To push a new image, run 'gcloud builds submit --project=cloud-eng-council --tag gcr.io/cloud-eng-council/make .'
```
morgante: We should probably just put in place a trigger to do this automatically on master.

helpers/check_format.sh
```diff
@@ -0,0 +1,69 @@
...
+    tools/labelmaker
+    tools/maven-archetype-dataflow
+    tools/ml-dataprep
+    tools/netblock-monitor
+    tools/site-verification-group-sync
+    tools/terraform-module-update-scanner
+    tools/bigquery-hive-external-table-loader
+)
+
+for FOLDER in $(find tools -maxdepth 1 -mindepth 1 -type d);
```
morgante: Any thoughts on excluding examples/ as well?

helpers/format.sh
```diff
@@ -0,0 +1,27 @@
...
+# style guidelines. This script is automatically called when running
+# "make fmt" at the root of the repository.
+#
+# NOTE: The files will be formatted in place.
+#
+# The following languages are currently supported:
+# - python (using yapf)
+
+# format all python files in place
+yapf -i -r --style google tools/**/*.py
```
morgante: Shouldn't this also reference the exclude list? Otherwise anyone running `make fmt` will generate a huge diff.

cloudbuild/Dockerfile
AdrienWalkowiak: yapf supports python 2 and 3 per the doc. I don't know if we need to install both for it to work.

cloudbuild/Dockerfile
AdrienWalkowiak: Is it possible to do this trigger only when the related files are touched on master?

helpers/format.sh
AdrienWalkowiak: true, will push some changes

helpers/check_format.sh
AdrienWalkowiak: the tests only apply to tools currently. I think it would be a good idea to add examples to it for futures contributions and exclude the current ones here too.

AdrienWalkowiak: done

helpers/format.sh
AdrienWalkowiak: done

cloudbuild/Dockerfile
morgante: Yes, you can configure cloud build triggers to trigger based on path.

AdrienWalkowiak: got it. I added a target to the makefile to build the new image when this new trigger runs

morgante: We should have a cloudbuild for this as well, no? Go ahead and configure the trigger if you can.

AdrienWalkowiak:
I created a trigger on this in the same project before this change:

https://pantheon.corp.google.com/cloud-build/triggers/06fa306b-d8ec-4c30-b14b-28c2fb06629b?project=cloud-eng-council&folder&organizationId

PR #322, opened 2019-09-24
Adding support for python3 and python3 files in make test helper script.

AdrienWalkowiak:
This makes sure make test works properly for both python2 and python3 files.

The caveat of this is that make fmt only works for python2 at this point.


AdrienWalkowiak:
Added support for python2 and python3 in make fmt helper script too.

Issue #325, opened 2019-10-02
Asset exporter tool - getting ImportError in GAE 


kojnp:
Just tried to set up from scratch in new project. Followed steps from readme
When running the cron job I get this

ImportError: cannot import name 'expr_pb2' from 'google.type' (/env/lib/python3.7/site-packages/google/type/__init__.py)
at <module> (/env/lib/python3.7/site-packages/google/iam/v1/policy_pb2.py:16)
at <module> (/env/lib/python3.7/site-packages/google/iam/v1/iam_policy_pb2.py:17)
at <module> (/env/lib/python3.7/site-packages/google/cloud/asset_v1/proto/assets_pb2.py:19)
at <module> (/env/lib/python3.7/site-packages/google/cloud/asset_v1/proto/asset_service_pb2.py:20)
at <module> (/env/lib/python3.7/site-packages/google/cloud/asset_v1/types.py:23)
at <module> (/env/lib/python3.7/site-packages/google/cloud/asset_v1/__init__.py:20)
at <module> (/srv/lib/asset_inventory/export.py:33)
at <module> (/srv/main.py:45)
at import_app (/env/lib/python3.7/site-packages/gunicorn/util.py:350)
at load_wsgiapp (/env/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py:41)
at load (/env/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py:52)
at wsgi (/env/lib/python3.7/site-packages/gunicorn/app/base.py:67)
at load_wsgi (/env/lib/python3.7/site-packages/gunicorn/workers/base.py:138)
at init_process (/env/lib/python3.7/site-packages/gunicorn/workers/base.py:129)
at init_process (/env/lib/python3.7/site-packages/gunicorn/workers/gthread.py:104)
at spawn_worker (/env/lib/python3.7/site-packages/gunicorn/arbiter.py:583)

bmenasha:
Sorry for the trouble. Let me try to recreate the issue.

bmenasha:
Thanks for the report. Fixed with the above pull request.

bmenasha:
should be resolved now. reopen if there is still a problem. thanks

bmenasha:
apparently i lack permissions to close.

PR #326, opened 2019-10-02
Update GAE requirements.txt to match Beam's versions.

bmenasha:
a version drift caused issue #325.
oauth2client also no longer supports file caching to prevent
warning from being logged add 'cache_discovery=False'


morgante:
@bmenasha Who is @vilmacs? If it's your account, could you change the commits to being made with your go/github account? Otherwise CLA bot will complain.

jaketf:
@bmenasha  I think there's a CLA issue cuz the commit author is not you. 
https://github.com/GoogleCloudPlatform/professional-services/pull/326/commits

bmenasha:
vilmacs is no more. only bmenasha@ here now.
thanks

PR #332, opened 2019-10-10
Adding Table Validator

nvurdien:


danieldeleo:
CLA warning can be ignored. All authors are googlers.

PR #336, opened 2019-10-11
Fixes for dataflow-data-generator and dataproc-persistent-history-server

jhambleton:
dataflow-data-generator:
- fast_avro and avro schema compatibility
- readme gs bucket location update
- minimum httplib2 requirement for avoiding ssl handshake error

dataproc-persistent-history-server:
- update prefix for mapreduce property in terraform


jhambleton:
@jaketf

PR #338, opened 2019-10-14
Removing Angular source files

kardiff18:
@danieldeleo removing angular source... I think I got it all but LMK if you see otherwise. Thanks!


danieldeleo:
I still see all the node_modules. I think you added them back when you fixed the merge conflict

PR #349, opened 2019-11-06
better documentation for slo-generator

victuos:
The links to the slo-generator examples were wrong, so I fixed them.


victuos:
@googlebot I signed it!

PR #362, opened 2019-11-30
New tool to generate GCS Bucket Insights

kardiff18:
This creates a tool that uses audit logs to understand usage patterns of GCS buckets for an entire org. Specifically, it looks to see how often/when objects within a bucket are read. Based on this, it then creates a new BigQuery table with these statistics as well as a recommended OLM for all objects inside that bucket.

An end user can use this table to then add in new OLM and better understand how downstream buckets are/are not being used.

Eng Council ID: 144777001


tools/gcs-usage-recommender/README.md
```diff
@@ -0,0 +1,247 @@
...
+    * <b>ADMIN_READ</b>: Entries for operations that read the configuration or 
+    metadata of a project, bucket, or object.
+    * <b>DATA_READ</b>: Entries for operations that read an object.
+    * <b>DATA_WRITE</b>: Entries for operations that create or modify an object.
+
+By default, only admin activity logs are enabled for organizations. However, data 
+access logs are not configured by default. This will require us to enable them in 
+the GCP console.
+
+To do this, go to the [console](https://pantheon.corp.google.com/iam-admin/audit?_ga=2.160269630.-2040617453.1540660549).
```
jaketf: remove pantheon link.

tools/gcs-usage-recommender/README.md
```diff
@@ -0,0 +1,247 @@
...
+ 
+* <b>Data Access logs</b>: Entries for operations that modify objects or read a 
+project, bucket, or object. There are several sub-types of data access logs:
+    * <b>ADMIN_READ</b>: Entries for operations that read the configuration or 
+    metadata of a project, bucket, or object.
+    * <b>DATA_READ</b>: Entries for operations that read an object.
+    * <b>DATA_WRITE</b>: Entries for operations that create or modify an object.
+
+By default, only admin activity logs are enabled for organizations. However, data 
+access logs are not configured by default. This will require us to enable them in
```
jaketf: "require" is a strong word.
can this really not be achieved w/ API Call / gsutil / terraform?
Is this because you're turning it on at org level?

tools/gcs-usage-recommender/README.md
```diff
@@ -0,0 +1,247 @@
...
+  --log-filter='resource.type="gcs_bucket" AND
+  (protoPayload.methodName:"storage.buckets.delete" OR
+   protoPayload.methodName:"storage.buckets.create" OR
+   protoPayload.methodName:"storage.buckets.get" OR
+  protoPayload.methodName:"storage.objects.get" OR
+  protoPayload.methodName:"storage.objects.delete" OR
+  protoPayload.methodName:"storage.objects.create")' \
+  --organization=${ORG_ID} --include-children
+````
+
```
jaketf: In the spirit of cost optimization, consider recommending a logging exclusion filter to prevent the $0.50/ GB stackdriver logging ingest for these additional audit logs.
https://cloud.google.com/logging/docs/export/

tools/gcs-usage-recommender/python/backfill.py
```diff
@@ -0,0 +1,124 @@
...
+    Args:
+        project_ids: List of strings holding project IDs
+        gcs_client: storage.Client object
+
+    Returns:
+        List of dictionaries mapping bucket-level metadata.
+    """
+    output_list = []
+    try:
+        for project_id in project_ids:
```
jaketf: In a large organization with 1000s of projects and potentially millions of buckets this could be slow have we run any benchmarks at scale?
 
Seems like this could greatly benefit from by parallelizing the `project_ids` loop. However, this might be premature optimization if we think this is truly a onetime activity where we could tolerate some latency.

tools/gcs-usage-recommender/python/backfill.py
```diff
@@ -0,0 +1,124 @@
...
+            bucket_list = list(gcs_client.list_buckets(project=project_id))
+            for bucket in bucket_list:
+                output_list.append({
```
jaketf: I'm having trouble here. Can you help me understand the significance of the values in this JSON?
Are these expected to be overwritten by BQ later?
Would it be safer to set las_read_timestamp to 30 days ago to avoid suggesting over aggressive OLM Policies?

tools/gcs-usage-recommender/README.md
```diff
@@ -0,0 +1,247 @@
...
+Currently, you are now collecting the raw audit log events representing GCS 
+objects/buckets. The missing link, however, is the events for the set of 
+buckets/objects which were created before the sink started generating these events.
+
+We’ll run a process to populate a new, separate table in our dataset with an 
+inventory of the currently existing buckets in your GCP organization to fill in 
+this missing link.
+
+This process only needs to be run once and would benefit from a low-latency 
+network location to GCP. In our example we will run it in the cloud shell.
```
jaketf: Cloud shell limited resources might not be the best place for listing this many buckets.

tools/gcs-usage-recommender/python/backfill.py
```diff
@@ -0,0 +1,124 @@
...
+import datetime
+import json
+import sys
+import logging
+from typing import Dict, List
+from google.cloud import resource_manager
+from google.cloud import storage
+
+
+def get_storage_client() -> storage.Client:
```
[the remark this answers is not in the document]
kardiff18: Yup, makes sense. I initially had some more GCS functionality in there, but ended up getting rid of it. But since it's only used once now, it's no longer needed. I can delete it.

tools/gcs-usage-recommender/python/backfill.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    output_list = []
+    try:
+        for project_id in project_ids:
+            bucket_list = list(gcs_client.list_buckets(project=project_id))
+            for bucket in bucket_list:
+                output_list.append({
+                    "bucket_name": bucket.name,
+                    "project_id": project_id,
+                    "last_read_timestamp": datetime.datetime(1970, 1, 1, 0, 0, 0).strftime("%Y-%m-%d %H:%M:%S"),
```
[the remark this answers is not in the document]
kardiff18: I was trying to default to essentially a null date, without using null explicitly. Do you have a suggestion of what to use instead? @jaketf

tools/gcs-usage-recommender/python/backfill.py
```diff
@@ -0,0 +1,124 @@
...
+        project_ids: List of strings holding project IDs
+        gcs_client: storage.Client object
+
+    Returns:
+        List of dictionaries mapping bucket-level metadata.
+    """
+    output_list = []
+    try:
+        for project_id in project_ids:
+            bucket_list = list(gcs_client.list_buckets(project=project_id))
```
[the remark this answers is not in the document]
kardiff18: Hmm.. so imap is only available in Python 2, and I wanted to be conducive to both Python 2/3 environments. I also ran this in our PSO org, and it didn't seem to have performance issues. But LMK what you think @jaketf

tools/gcs-usage-recommender/README.md
kardiff18: I filter only on the specific actions for GCS it requires (so I exclude out some ops, for things like service accounts, etc.) via the method names needed. Is there anything else that you'd recommend @jaketf ?

tools/gcs-usage-recommender/README.md
kardiff18: Yes, although since it's a one-off run, I would argue that the latency may be tolerated this time.

tools/gcs-usage-recommender/python/backfill.py
kardiff18: I ran it on our org, as well as a test GCS org from an SA. Both of which performed surprisingly well. Also, since it is a onetime activity, the small amount of latency that there is can probably be tolerated.

tools/gcs-usage-recommender/python/backfill.py
kardiff18: Yes, these are meant to capture all buckets that existed in an org prior to enabling audit logs. Because of this, they have default dummy values, that are overwritten once the bucket appears in the audit logs.

Open to suggestions on last_read_timestamp, although I would think that a more obvious dummy value would perhaps be better to show that it truly has not been accessed. A 30-days-ago timestamp may lead someone to incorrectly believed that it was read, when in fact, it maybe hasn't been accessed in a year. In this case, the OLM is blank regardless, so I don't think aggressive OLM policies will be in effect.

tools/gcs-usage-recommender/python/backfill.py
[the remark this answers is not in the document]
jaketf: I think a null date is more honest (see other comment thread).
This comment was just about favoring named constants over hard coded for readability. 
See [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))

tools/gcs-usage-recommender/python/backfill.py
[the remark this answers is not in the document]
jaketf: ah TIL imap in python2 -> map in python3 ([source](https://stackoverflow.com/questions/30271712/i-cant-find-imap-in-itertools-in-python)), thanks!
I think we should just leave this as is.

I think the way to do this preserving the iterator would be to use `map` and `chain.from_iterable` but it becomes much less readable and based on some testing we'd have to scale tot 100s of millions of buckets before there'd be real danger of memory issues on cloud shell / n1-standard1. I think my bias for iterator comes from object listing scenarios where millions of objects is more likely and where the benefit of lazily evaluated iterators can be safer.

If you're curious this would work in python2/3 and benefit from iterator if python3.
```python3
bucket_itr = itertools.chain.from_iterable(
    map(lambda project_id: gcs_client.list_buckets(project=project_id), 
             project_ids))
```

tools/gcs-usage-recommender/README.md
jaketf: What you're specifying here is a filter for the logging export (in this case to BQ).

However, as is all these GCS logs will be ingested into stackdriver logging in addition to BQ (which can add up). If the primary way the organization will look at GCS logs is through BQ (not Stack driver logging search / monitoring / alerting), then you should also exclude these logs from stackdriver logging so you're not paying $0.50/GB ingest for logs you're already ingesting to BQ (for much cheaper).

To do this you'd have to specify a separate filter to avoid these logs getting ingested into Stackdriver Logging via an [Exclusion](https://cloud.google.com/logging/docs/exclusions)
I would recommend mentioning exclusions as an optional step as there are valid reasons a user might want these logs in stackdriver logging as well (namely alerting).
I only brought this up because I think the spirit of this asset is cost savings and it's a good opportunity to remind customers that they can also save on their stackdriver bill by only ingesting to stackdriver logging what is actually giving them value in stackdriver (e.g. logs based metrics / alterting).

tools/gcs-usage-recommender/python/backfill.py
jaketf: Agreed.
Out of curiosity, do you have rough number of projects / buckets and time your benchmark took? Is this in line with the project / bucket cardinality of our larger customers?

tools/gcs-usage-recommender/python/backfill.py
jaketf: This is a tricky one and you're right to have concern about misleading the user on last read date. I'd argue that the same could be said about setting the dummy date so far back misleading them to believe an object hasn't been read for years when perhaps it has been read in the last 90 days.
If we set 30 days ago date we'll recommend near line for all the backfill data (that hasn't appeared in the audit logs). Then after 60 days we'll either recommend going to cold line or not based on actual information we observed.   
I think we really want to avoid a scenario where we suggest moving many objects to cold line storage and then have to convert many of them back. Changing storage class of many objects can take GCS a long time and affects read performance. This can be very dramatic for customers who are moving stuff to cold line and then simultaneously trying to run a yearly report on those same objects.

Curious what you'd think about a null date?
It might be annoying to deal with but I think it avoids this pickle of picking a misleading default.

tools/gcs-usage-recommender/README.md
kardiff18: Ah, okay sorry did not understand, very cool! Thanks LGTM

tools/gcs-usage-recommender/python/backfill.py
kardiff18: Hmm.. agreed on the performance of changing storage classes. However, this defaults OLM to be blank, btw, not coldline or nearline, so whichever dummy value we decide on affects just a user's understanding of the bucket not an OLM. The OLM is generated only when it appears in the audit logs.

1. My only concern about doing a null date was because I wanted the schema to default to datetime type, but I just realized that I already set up the schema file. I agree and will change to null :)

tools/gcs-usage-recommender/python/backfill.py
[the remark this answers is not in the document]
kardiff18: Done -- see other thread

PR #366, opened 2019-12-06
Added in sample audio file for existing solution

kardiff18:
Added in sample audio file to use for the solution as the speech API relies on specific encoding (both .FLAC format and mono with 1 audio channel.) The recording was produced by an instructional designer at Google, by reading a solution published [here](https://cloud.google.com/solutions/machine-learning/categorizing-audio-files-using-ml)


jaketf:
I don't usually see audio files checked into github. 
Would this be more appropriate to host in a publicly readable GCS bucket and refer to it in the docs for this asset?

jaketf:
Lgtm

PR #371, opened 2019-12-17
Added append argument to use WRITE_APPEND option

burcuidikut:
When we call run_query function with overwrite=False argument, query is executed as WRITE_EMPTY (default append value=False, default overwrite value=True),
when we call run_query function with overwrite=True argument, query is executed as WRITE_TRUNCATE,
when we call run_query function without overwrite argument, query is executed as WRITE_TRUNCATE again, in any case we cannot use append (WRITE-APPEND) option. For this reason, we have to add append argument to run_query function definition, run_queries function definition and job variable assignment in run_query function, or we can change if statement for overwrite/append in create_job_config function.


jaketf:
@burcuidikut FYI checkout ox-bqpipeline it's a fork of this with a few more features https://github.com/openx/ox-bqpipeline

PR #390, opened 2020-01-10
Cloud Composer CI/CD example

tims:
Example that uses Cloud build to test airflow DAGs and deploy them to Cloud Composer.

When deploying cloud composer environments, people often ask, "How do I test my DAGs".
The answer is CI/CD

* DAG integrity / validation tests (check it load correctly).
* DAG execution tests (run the complete dag)

Then deploy.

I've prepared an example of this using cloud build.

PSO Eng Bug Id = 147466944

### DIFF
--- a/examples/cloud-composer-cicd/cloudbuild.yaml
+++ b/examples/cloud-composer-cicd/cloudbuild.yaml
@@ -1,0 +1,81 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  id: Pull docker cache
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+   docker pull gcr.io/$PROJECT_ID/airflow-dags-builder:latest || exit 0
+- name: gcr.io/cloud-builders/docker
+  id: Build Airflow DAGs Builder
+  args: [
+      'build',
+      '-t', 'gcr.io/$PROJECT_ID/airflow-dags-builder',
+      '--cache-from', 'gcr.io/$PROJECT_ID/airflow-dags-builder:latest',
+      './'
+    ]
+  dir: ${_DIRECTORY}
+- name: 'gcr.io/$PROJECT_ID/airflow-dags-builder'
+  id: Validation Test
+  # Validate the integrity of the DAG files.
+  entrypoint: python
+  env:
+  - AIRFLOW__CORE__DAGS_FOLDER=/workspace/${_DIRECTORY}/dags
+  args:
+  - -m
+  - unittest
+  - tests/dag_integrity_test.py
+  dir: ${_DIRECTORY}
+- name: 'gcr.io/$PROJECT_ID/airflow-dags-builder'
+  # Actually run some DAGs with test data, this type of test is less common and requires
+  # more setup, creating test data sets, ensuring your task IO is configured
+  # via airflow Variables, etc.
+  # To succeed, these need to complete before the cloudbuild timeout.
+  id: End to End Test
+  entrypoint: bash
+  env:
+  - AIRFLOW__CORE__DAGS_FOLDER=/workspace/${_DIRECTORY}/dags
+  dir: ${_DIRECTORY}
+  # dir: examples/cloud-composer-cicd
+  args:
+  - -c
+  - |
+    pwd
+    ls -l
+    # Init an empty airflow metadata db
+    airflow initdb
+    # Initialise variables to configure airflow DAGs
+    airflow variables -s input gs://test-input-bucket/
+    airflow variables -s output gs://test-output-bucket/
+    # Run a backfill of our DAG for a specific date, add more dags here if required.
+    airflow backfill echo_and_wait --start_date=2020-01-03 --end_date=2020-01-03
+    ret=$?
+    # Cat the task logs so we can see when there was any error
+    find /workspace/airflow/logs/ -type f -name *.log | sort | xargs cat
+    # Fail if the backfill failed.
+    exit $ret
+- name: gcr.io/cloud-builders/gsutil
+  # Deploy the DAGs to your composer environment DAGs GCS folder
+  id: Deploy DAGs
+  args:
+  - -m
+  - rsync
+  - -r
+  - -c
+  - -x
+  - .*\.pyc|airflow_monitoring.py
+  - /workspace/${_DIRECTORY}/dags/
+  - ${_DEPLOY_DAGS_LOCATION}
+  dir: ${_DIRECTORY}
+images: ['gcr.io/$PROJECT_ID/airflow-dags-builder:latest']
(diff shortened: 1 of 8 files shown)


examples/cloud-composer-cicd/cloudbuild.yaml
```diff
@@ -0,0 +1,81 @@
...
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
```
henryken: Just FYI, I have a pull request extracting out the first two steps into its own cloud builder in the cloud builder community to avoid running it for every build.
Maybe we can incorporate it once it is merged.
https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/346

examples/cloud-composer-cicd/cloudbuild.yaml
```diff
@@ -0,0 +1,81 @@
...
+    # Initialise variables to configure airflow DAGs
+    airflow variables -s input gs://test-input-bucket/
+    airflow variables -s output gs://test-output-bucket/
+    # Run a backfill of our DAG for a specific date, add more dags here if required.
+    airflow backfill echo_and_wait --start_date=2020-01-03 --end_date=2020-01-03
+    ret=$?
+    # Cat the task logs so we can see when there was any error
+    find /workspace/airflow/logs/ -type f -name *.log | sort | xargs cat
+    # Fail if the backfill failed.
+    exit $ret
```
jaketf: This should be taken out into it's own script.
This also notably will cause red-herring test failures because it won't handle:
- if the customer has any custom conn_ids
- if the customer is using crypto / fernet key
- if the customer is also deploying custom plugins.

PTAL https://source.cloud.google.com/datapipelines-ci/datapipelines-ci/+/master:composer/cloudbuild/bin/run_tests.sh

examples/cloud-composer-cicd/cloudbuild.yaml
```diff
@@ -0,0 +1,81 @@
...
+- name: gcr.io/cloud-builders/gsutil
+  # Deploy the DAGs to your composer environment DAGs GCS folder
+  id: Deploy DAGs
+  args:
+  - -m
```
jaketf: nit: I have had really flaky issues with deleting DAGs w/ composer when paused on creation is not manually configured to `True`. I also consider this a best practice in case a dag is manually copied here by accident and shouldn't actually be scheduled.

If this config is set this script won't handle the dag unpausing. 

Note, (though not necessary) I've seen a desire to have a config that only deploys certain dags. Sometimes, users may need to run multiple deploys to sequence their updating of DAGs due to inter-dag dependencies or they have some utility temporary dags they want to deploy run and undeploy w/o removing the dag from their codebase. 
I've solved this with a running_dags.txt

examples/cloud-composer-cicd/cloudbuild.yaml
tims: I don't have access to that repo, permission denied

tims:
> I've been working in https://source.cloud.google.com/datapipelines-ci/datapipelines-ci/+/master
I don't have access to that repo, permission denied

examples/cloud-composer-cicd/cloudbuild.yaml
tims: I don't have access to that repo, permission denied

examples/cloud-composer-cicd/cloudbuild.yaml
tims: That should be an environment wide setting no?

examples/cloud-composer-cicd/cloudbuild.yaml
tims: I thought it might be more readable in the same file. And since this only tests one DAG this way, it's more of an example than a best practice. This is just to demo how you might override variables before creating and executing a DAG run. 

I definitely agree a more complete solution would be great. But it would be good to have something simple like this to link to until then. We can change to link to that when it's ready.

examples/cloud-composer-cicd/cloudbuild.yaml
tims: My preference would be to set the environment wide setting to set all new dags as paused on creation.

If a dag is updated, I don't think it should ever automatically unpause. I'd much prefer the user has to do that once manually.

But this is very user and environment specific. I don't think CI/CD should be particularly opinionated about it.

examples/cloud-composer-cicd/cloudbuild.yaml
tims: I'd suggest we merge this for now, and then come up with the better solution next. I'd suggest we're much better off having velocity and revising than waiting for perfect.

tims: Also, I consider the main benefit to this example to be just the syntax checking, it covers the most immediate need for customers. People tend to do very different things for end to end tests, and most don't, they just see what happens in prod. So it's more of a starting point and helps explain how simple doing things with cloud build is.

jaketf: i opened a PR a while back for the cicd for data pipelines solution that didn't get much traction is very similar to what I'm working on.
You could link to my fork: https://github.com/jaketf/ci-cd-for-data-processing-workflow
I think that handling  plugins / connections is a must even for a simple example.
IMHO cloud build should be like a main function: very small high level readable steps, all implementation details like initdb, variables, etc should be burried in scripts.

examples/cloud-composer-cicd/cloudbuild.yaml
jaketf: that's fair...

jaketf:
@tims please try to access again.
If you can take this run tests scripts and use it or something similar to handle plugins, variables, connections, I'd be much more comfortable merging this.

jaketf:
@tims 
LMK if you have any questions about using the run tests script I've linked. 

FYI this is where I'm hoping to open source this as a refactor of the public solution. 
https://github.com/GoogleCloudPlatform/ci-cd-for-data-processing-workflow/pull/2

tims:
@jaketf Unfortunately I think there are a few issues with the run_tests script you linked to, it's more generic but not generic enough.

It takes connections from an existing composer environment. But testing should probably not use the same connections as a composer env. I think they should be provided by the build, if any. It also assumes HTTP connections, which is not ideal.

It also assumes the variables are in a specifically named local file, presumably committed, if we're going to make it more generic than inlining in the cloudbuild.yaml, these should be configurable.

It expects a plugins dir, which a simple example does not require. Similarly it is expects bigquery/sql dir.

I still think we should merge as is, and link to yours as the advanced solution when it's ready.

examples/cloud-composer-cicd/cloudbuild.yaml
```diff
@@ -0,0 +1,81 @@
...
+    airflow variables -s input gs://test-input-bucket/
+    airflow variables -s output gs://test-output-bucket/
+    # Run a backfill of our DAG for a specific date, add more dags here if required.
+    airflow backfill echo_and_wait --start_date=2020-01-03 --end_date=2020-01-03
+    ret=$?
+    # Cat the task logs so we can see when there was any error
+    find /workspace/airflow/logs/ -type f -name *.log | sort | xargs cat
+    # Fail if the backfill failed.
+    exit $ret
+- name: gcr.io/cloud-builders/gsutil
```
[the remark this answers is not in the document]
tims: This would requires a composer environment to be running to run the tests, which given this is a lightweight example I didn't want to require. Would also require extra config of the composer env.

tims:
@jaketf we can definitely open up a new PR on this when your is ready to converge. This one should be kept as simple as possible as an illustration for people are rolling their own CI/CD and (some customers are even only using airflow, not composer). And we can put the more complete solution in your repo.

jaketf:
@tims makes sense glad you go this merged. thank you that's really valuable feedback on the run tests script. I think there's huge value in making it a bit more configurable / modular. I've migrated this to a PR in the open it will be a while til we can rewrite the solutions page but I'd appreciate any review comments like this! https://github.com/Google loud platform/ci-cd-for-data-processing-workflow/pull/2

PR #400, opened 2020-02-02
Update ReadMe.md

adelanre:
- Added Destination to log sink create.
- Add missing `\` to the gcloud function deploy.
- Add `LOG_EXPORT_NAME` var


adelanre:
@googlebot I signed it!

examples/dlp/cloud_function_example/README.md
```diff
@@ -62,7 +64,7 @@ Create the export.
 export PROJECT_ID=$(gcloud config get-value project)
 
 gcloud logging sinks create $LOG_EXPORT_NAME \
-  pubsub.googleapis.com/projects/$PROJECT_ID/topics/$LOG_EXPORT_TOPIC_NAME \
+  $LOG_EXPORT_TOPIC_NAME pubsub.googleapis.com/projects/$PROJECT_ID/topics/$LOG_EXPORT_TOPIC_NAME \
```
tims: I think you added the $LOG_EXPORT_TOPIC_NAME by mistake?

Docs say it should be `gcloud logging sinks create SINK_NAME DESTINATION`. 
The destination should only be the fully qualified destination topic?

adelanre: @tims That's what I thought as well until I tried to run the command a second time. 
I get this error when I tried to use DESTINATION a second time ```is the subject of a conflict: Sink DESTINATION already exists```

image

PR #404, opened 2020-02-10
Fix arithmetics in tools/gce-quota-sync

miklosn:
Fix division to use floating point arithmetics (it results in 0.0 otherwise).  Also add sending in the usage value.


tools/gce-quota-sync/gce-quota-sync.py
```diff
@@ -132,15 +132,15 @@ def _quota_to_series(project, region, quota):
     region: set in converted time series labels
     quota: quota object received from the GCE API
   """
-  labels = dict((k, str(v)) for k, v in quota.items() if k != 'usage')
+  labels = dict((k, str(v)) for k, v in quota.items())
```
jaketf: why was usage originally omitted / now included ?

tools/gce-quota-sync/gce-quota-sync.py
```diff
@@ -132,15 +132,15 @@ def _quota_to_series(project, region, quota):
...
+
   labels['project'] = project
   labels['region'] = region
   try:
-    value = quota['usage'] / quota['limit']
+    value = quota['usage'] / float(quota['limit'])
```
jaketf: IIUC python3 this [casting as float is unecessary](https://www.python.org/dev/peps/pep-0238/) when using division operator.

tools/gce-quota-sync/gce-quota-sync.py
miklosn: I have no idea. It was included in the examples as a feature and i definitely found the number to be useful in stackdriver.

tools/gce-quota-sync/gce-quota-sync.py
miklosn: Exactly :) My client ran it on python2, figured it wouldn't hurt.

jaketf: This is fine for backwards compatibility but your client should get w/ the times :)

miklosn: Haha yes.

PR #407, opened 2020-02-19
gce-quota-sync: Add linux-headers to alpine pkgs installed in Docker & README update

pehowell:
To build the python gRPC library, Alpine needs the linux headers installed. This PR:

- adds `linux-headers` to the list of alpine packages installed and later removed during Docker build
- updates README to reflect the correct [project option](https://github.com/GoogleCloudPlatform/professional-services/blob/433da74fa5665dff9bb368133b46c384731eb0e2/tools/gce-quota-sync/gce-quota-sync.py#L146) that must be given


pehowell:
Thanks for the approval @jaketf, anything I need to do on my end to merge?

PR #410, opened 2020-02-21
Update BQ Benchmark Code to Python3

annarudy:
Updated all files in bq_file_load_benchmark to Python3


examples/bq_file_load_benchmark/generic_benchmark_tools/table_util.py
```diff
@@ -185,7 +182,14 @@ def get_column_types(self):
 
         schema = self.table.schema
         field_types = [field.field_type for field in schema]
-        field_type_counts = collections.Counter(field_types)
+        field_types.sort()
```
jaketf: Why do you import `collections.Counter` but re-implement t here?
Is this not equivalent of 
```suggestion
        field_type_counts = Counter(field_types)
```

examples/bq_file_load_benchmark/benchmark_tools/benchmark_result_util.py
```diff
@@ -1,284 +0,0 @@
-# Copyright 2018 Google Inc.
```
jaketf: Is it intended that this file  was deleted as part of this python3 migration?

examples/bq_file_load_benchmark/generic_benchmark_tools/table_util.py
annarudy: I forgot to take out the collections import. i figured it made as much sense to just do it manually.

examples/bq_file_load_benchmark/benchmark_tools/benchmark_result_util.py
annarudy: nope that should have been done in an earlier PR. I just noticed it here and deleted it.

jaketf: This is going to remove this file from master if we merge.
Is that what you mean to do?

examples/bq_file_load_benchmark/generic_benchmark_tools/table_util.py
jaketf: IMHO this is not the right choice.

We should not reimplenent key features of the language because we _can_. Often times this is prone to missing failure cases that others have already protected against in the core feature.

Unless you can prove that the core feature doesn't meet your use case and that your implementation addresses the gap without regression (by adding extensive unittests).
Use `collections.Counter` for 
- maintainability: no need to add additional tests 
- error handling: even outside this loop `Counter` will give better guarantees e.g.  returning 0 instead of raising `KeyError` when requesting the count for a key that is not part of this counter object.

examples/bq_file_load_benchmark/benchmark_tools/benchmark_result_util.py
annarudy: yes

examples/bq_file_load_benchmark/generic_benchmark_tools/table_util.py
annarudy: ok i reverted it back

annarudy:
@jaketf changes have been made/addressed

PR #414, opened 2020-02-25
Replace constant with instance variable in Fixity example

zefdelgadillo:
A user of this project noticed an issue where the Fixity timestamp constant persisted across various runs of the Cloud Function, which caused an issue with the results. 

The tool expects that a new timestamp is generated for each new invocation of the Cloud Function, so using a constant is the wrong way to set the date. This proposed change resolves that, using a new timestamp for each run.

### DIFF
--- a/examples/gcs-fixity-function/src/main.py
+++ b/examples/gcs-fixity-function/src/main.py
@@ -24,7 +24,6 @@ import json
 from datetime import datetime
 from google.cloud import storage, bigquery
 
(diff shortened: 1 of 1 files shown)


zefdelgadillo:
Sorry about that @jaketf, slight hiccup in the naming there. Should I update the commit message too or is this ok to merge?

jaketf:
@zefdelgadillo nah I'll just correct it in the squashed commit

PR #418, opened 2020-03-02
GCS IAM Policy Util - manage Data Lake folder security via CI/CD

jasonmar:
This is a command-line utility that automates the application of GCS IAM policies with IAM conditions.

It's fully configuration driven and greatly simplifies management of IAM policies at scale.
Compared to the Cloud Console or gsutil where manual policy creation is time consuming, users can manage thousands of policies easily via configuration with this tool.

_See README.md for more details._

**Example**

Running with this minimal configuration:

```
[gs://mybucket/hive/warehouse/mytable]
roles/storage.objectAdmin serviceAccount:data-admin@dataproject.iam.gserviceaccount.com
roles/storage.objectViewer serviceAccount:data-viewer@dataproject.iam.gserviceaccount.com
```

Would create and apply the following IAM policy:

```
{
  "bindings": [
    {
      "condition": {
        "expression": "resource.name.startsWith(\"projects/_/buckets/mybucket/objects/hive/warehouse/mytable\")"
      },
      "members": [
        "serviceAccount:data-admin@dataproject.iam.gserviceaccount.com"
      ],
      "role": "roles/storage.objectAdmin"
    },
    {
      "condition": {
        "expression": "resource.name.startsWith(\"projects/_/buckets/mybucket/objects/hive/warehouse/mytable\")"
      },
      "members": [
        "serviceAccount:data-viewer@dataproject.iam.gserviceaccount.com"
      ],
      "role": "roles/storage.objectViewer"
    },
    {
      "members": [
        "serviceAccount:gcs-iam@adminproject.iam.gserviceaccount.com"
      ],
      "role": "roles/storage.admin"
    }
  ],
  "etag": "CBk="
}
```


jaketf:
@jasonmar for my understanding why would a customer learn your non-standard config spec to use this tool, over using something like this: [google_storage_bucket_iam_member](https://www.terraform.io/docs/providers/google/r/storage_bucket_iam.html#google_storage_bucket_iam_policy-1) with a condition(s) block?

Are there features you expose with this tool that make it superior? 
Is the use case just a customer that doesn't use terraform? 

I feel like it would be more accessible for users and more maintainable for us if we created a terraform module for this use case.

However, I bet I'm missing something glaring!

jasonmar:
This is the most compact representation possible.
With terraform, if you have a thousand buckets all needing the same policy, (for example, in a data lake with one bucket per dataset) there will be a lot of duplication. With this tool, you could define a group of buckets and apply the policy to the entire group.

morgante:
> Terraform is great, but if you have a thousand buckets all needing the same policy, (for example, in a data lake with one bucket per dataset) there will be a lot of duplication.

This could certainly be handled succinctly in Terraform through a module or for_each.

jaketf:
I don't thinks it's the concision as much as the dependency on a scala app when many customer already use terraform for managing IAM like this (because they can link dynamic to references to SAs / buckets also being provisioned in the same apply or in another workspace etc).

In this example you'd have to sed replace on your json path to refer to new SA / buckets in prod than dev. In terraform we get that string interpolation for free. 

We could have a module that takes dataset_buckets, admins, viewers if the use case is same permissions many buckets and do something like this:
```hcl
// In a module this could be user input
locals {
  dataset_buckets = {
    dataset_0 = {
      bucket           = "bucket0"
      prefix           = "hive/warehhouse/dataset0"
      explicit_viewers = ["user:joe@example.com"]
      explicit_admins  = ["user:bob@example.com"]
    },
    dataset_1 = {
      bucket           = "bucket2"
      prefix           = "hive/warehhouse/dataset2"
      explicit_viewers = ["user:joe@example.com"]
      explicit_admins  = ["user:bob@example.com"]
    }
  }
  global_admins = [
    "serviceAccount:data-admin@dataproject.iam.gserviceaccount.com",
  ]
  global_viewers = [
    "serviceAccount:data-viewer@dataproject.iam.gserviceaccount.com",
  ]
}

provider "google-beta" {}

resource "google_storage_bucket_iam_binding" "data_admins" {
  for_each = local.dataset_buckets

  bucket   = each.value["bucket"]
  provider = google-beta
  role     = "roles/storage.admin"
  members  = concat(local.global_admins, each.value["explicit_admins"])

  condition {
    title       = "data_admin_with_prefix_${each.key}"
    description = "Expiring at midnight of 2019-12-31"
    expression  = "resource.name.startsWith(\"${each.value["prefix"]}\")"
  }
}

resource "google_storage_bucket_iam_binding" "data_viewers" {
  for_each = local.dataset_buckets
  provider = google-beta

  bucket  = each.value.bucket
  role    = "roles/storage.objectViewer"
  members = concat(local.global_viewers, each.value["explicit_viewers"])

  condition {
    title       = "data_admin_with_prefix_${each.key}"
    description = "Expiring at midnight of 2019-12-31"
    expression  = "resource.name.startsWith(\"${each.value["prefix"]}\")"
  }
}
```
plan: 
```
Terraform will perform the following actions:

  # google_storage_bucket_iam_binding.data_admins["dataset_0"] will be created
  + resource "google_storage_bucket_iam_binding" "data_admins" {
      + bucket  = "bucket0"
      + etag    = (known after apply)
      + id      = (known after apply)
      + members = [
          + "serviceAccount:data-admin@dataproject.iam.gserviceaccount.com",
          + "user:bob@example.com",
        ]
      + role    = "roles/storage.admin"

      + condition {
          + description = "Expiring at midnight of 2019-12-31"
          + expression  = "resource.name.startsWith(\"hive/warehhouse/dataset0\")"
          + title       = "data_admin_with_prefix_dataset_0"
        }
    }

  # google_storage_bucket_iam_binding.data_admins["dataset_1"] will be created
  + resource "google_storage_bucket_iam_binding" "data_admins" {
      + bucket  = "bucket2"
      + etag    = (known after apply)
      + id      = (known after apply)
      + members = [
          + "serviceAccount:data-admin@dataproject.iam.gserviceaccount.com",
          + "user:bob@example.com",
        ]
      + role    = "roles/storage.admin"

      + condition {
          + description = "Expiring at midnight of 2019-12-31"
          + expression  = "resource.name.startsWith(\"hive/warehhouse/dataset2\")"
          + title       = "data_admin_with_prefix_dataset_1"
        }
    }

  # google_storage_bucket_iam_binding.data_viewers["dataset_0"] will be created
  + resource "google_storage_bucket_iam_binding" "data_viewers" {
      + bucket  = "bucket0"
      + etag    = (known after apply)
      + id      = (known after apply)
      + members = [
          + "serviceAccount:data-viewer@dataproject.iam.gserviceaccount.com",
          + "user:joe@example.com",
        ]
      + role    = "roles/storage.objectViewer"

      + condition {
          + description = "Expiring at midnight of 2019-12-31"
          + expression  = "resource.name.startsWith(\"hive/warehhouse/dataset0\")"
          + title       = "data_admin_with_prefix_dataset_0"
        }
    }

  # google_storage_bucket_iam_binding.data_viewers["dataset_1"] will be created
  + resource "google_storage_bucket_iam_binding" "data_viewers" {
      + bucket  = "bucket2"
      + etag    = (known after apply)
      + id      = (known after apply)
      + members = [
          + "serviceAccount:data-viewer@dataproject.iam.gserviceaccount.com",
          + "user:joe@example.com",
        ]
      + role    = "roles/storage.objectViewer"

      + condition {
          + description = "Expiring at midnight of 2019-12-31"
          + expression  = "resource.name.startsWith(\"hive/warehhouse/dataset2\")"
          + title       = "data_admin_with_prefix_dataset_1"
        }
    }

```

jasonmar:
The other thing this does that terraform may not do is verify that the
service accounts are in your org.
It does this by querying the resource manager API.

The terraform for_each looks good and we should definitely share examples
of that.

>

jaketf:
> The other thing this does that terraform may not do is verify that the service accounts are in your org. It does this by querying the resource manager API. 

That is definitely a nice feature.
Yeah TF won't do that check out of the box.
Seems like if a customer cares about this would be a reasonable org policy restriction FR (I couldn't find one that does restrict iam permission / bucket access to w/ in org at first glance) so they could benefit from that restriction beyond the scope of this tool.
Though this might be because the best way to avoid ex-filtration out of the org that is create VPC-SC perimeter around the org?

ryanmcdowell:
> > The other thing this does that terraform may not do is verify that the service accounts are in your org. It does this by querying the resource manager API.
> 
> That is definitely a nice feature.
> Yeah TF won't do that check out of the box.
> Seems like if a customer cares about this would be a reasonable org policy restriction FR (I couldn't find one that does restrict iam permission / bucket access to w/ in org at first glance) so they could benefit from that restriction beyond the scope of this tool.
> Though this might be because the best way to avoid ex-filtration out of the org that is create VPC-SC perimeter around the org?

You can use domain restricted sharing to restrict sharing to a specific org and the service accounts associated with that org. This is an org policy which can also be defined in Terraform.

Realistically, to maintain an acceptable security posture, every GCP org should be using Domain Restricted Sharing and VPC-SC.

Docs
https://cloud.google.com/resource-manager/docs/organization-policy/restricting-domains

Service Account Design
go/iam-drs-sa

jasonmar:
Agreed, validating service account domain is redundant if DRS is enabled.

jasonmar:
This is targeted at security engineers who want to build their own custom tooling and need something terraform doesn't provide.
More specifically, it's targeted at data lake administrators who want to replicate Apache Ranger HDFS Policies in GCS without Apache Ranger.
It looks like it's possible to manage IAM policies with Terraform, but when I looked I couldn't find any examples of using the Java SDK to set IAM conditions.
If you think everyone should just use Terraform and there's no benefit to giving users code they can modify to fit their needs, feel free to say so and close this PR.

jasonmar:
The json config you're referencing is dynamically generated and only shown in the readme to expose the actual policy that gets applied - the user never actually sees it or modifies it as json. Only the 3 line config in the example is provided by the user.

> This example is billed as CI/CD but you'd have to sed replace on your json config (which could be error prone) to refer to new SA / buckets in prod than dev. In terraform we get that string interpolation / variablization of our configs for free.

I don't understand where you're thinking sed or string interpolation would be used. If there are distinct sets of buckets, you'd have two distinct policies. If you had one policy common to all, you could just create a zone called `all_buckets` that includes both prod and dev buckets.

```
[prod_buckets]
roles/storage.objectAdmin prod_admins
roles/storage.objectViewer prod_viewers

[dev_buckets]
roles/storage.objectAdmin prod_admins
roles/storage.objectAdmin dev_all
```

jasonmar:
I understand terraform reasonably well but these examples are not easy for me to parse at a glance.

If you were reviewing this diff, there's so much distracting text that it's hard to tell what's actually being granted:
`roles/storage.admin` on `dataset_buckets` to `explicit_admins`

Aside from brackets this is all that you'd need to provide if using this tool. It's just so much easier to review.

Compare this:

```
resource "google_storage_bucket_iam_binding" "data_admins" {
  for_each = local.dataset_buckets

  bucket   = each.value["bucket"]
  provider = google-beta
  role     = "roles/storage.admin"
  members  = concat(local.global_admins, each.value["explicit_admins"])

  condition {
    title       = "data_admin_with_prefix_${each.key}"
    description = "Expiring at midnight of 2019-12-31"
    expression  = "resource.name.startsWith(\"${each.value["prefix"]}\")"
  }
}
```

To this:

```
[dataset_buckets]
roles/storage.admin explicit_admins
```

What percentage do you think would choose the terraform example over the config format for this tool if you were to ask 100 random policy reviewers what they'd prefer and what's less error prone for them to review?

jasonmar:
The whole point of this is to make it so that admins can read the config file and understand what policies  are deployed. In my opinion the terraform examples do not achieve this, although terraform does make it easy to apply policies with CI/CD.

jaketf:
My apologies for misunderstanding that the json config is generated, I misunderstood that from the PR description.

I'm not trying to say everyone should use terraform, I'm trying to understand the audience for this tool and scenarios where this is a better fit so I could understand when to recommend which approach to a customer.

I think it's valid reasoning for a customer to say "we have security engineers with java expertise and processes around using java for managing policies that way" and am supportive of having examples for different audiences for similar challenges will help us reach more customers.

Regarding separate policies per environment, most organizations I've seen have dev, qa / staging, prod and the qa / staging environment should be be prod identical for all policies as it's purpose is to catch issues promoting policies (e.g. breaking an app by promoting an overly restrictive policy). So there is most value in interpolation when wanting to say "we validated this exact set of policies on this bucket naming convention in staging now push that to prod"

jasonmar:
The way they could achieve this is by having two bucket config files, one for dev and one for prod.

## zones_prod.conf
```
[dataset_buckets]
dataset-prod prefix
```

## zones_dev.conf
```
[dataset_buckets]
dataset-dev prefix
```

## policy.conf
```
[dataset_buckets]
roles/storage.admin explicit_admins
```

So they'd just run the tool pointing to the environment specific zones file by adding this argument to their command line:
`--zones "zones_${ENV}".conf`
where ENV is set to either `dev` or `prod`

jasonmar:
Thanks for providing these examples, btw. I'll add notes from the discussion to the readme before we merge.

jasonmar:
Regarding the bucket naming convention, I've seen some users have buckets which do not follow a naming convention, so string interpolation can't be used. If using hashes as bucket names, being able to use aliases makes things a lot easier for reviewers. How to verify the intended purpose of a bucket with a random name is a different story, but at least aliases make that naming convention more viable assuming the validation is done at review time before allowing configuration lines into the repository.

PR #420, opened 2020-03-05
Billing report tool

katze120:
This tool generates reports of Cloud Billing data from a BigQuery dataset, for a given month in csv format.

Main features include:
- Generation of multiple reports at a time with customizable report title and columns
- Support having label keys as report columns (which is usually challenging with pure SQL queries due to potential double counting)

PSO Eng Council ticket ID: 150855522


katze120:
@soellman can this PR be reviewed please?

PR #422, opened 2020-03-07
Feature/cdap lineage exporter

jaketf:
This tool allows users to easily export all the Dataset and Streams lineage from a Data Fusion instance to GCS.

PR #424, opened 2020-03-11
add integration tests. use google-api-core instead of google-cloud-core

tswast:
Improvements to https://github.com/GoogleCloudPlatform/professional-services/pull/422

### DIFF
--- a/tools/cdf-lineage-exporter/tests/unit/test_client.py
+++ b/tools/cdf-lineage-exporter/tests/unit/test_client.py
@@ -1,0 +1,47 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest import mock
+
+import pytest
+import requests
+
+
+@pytest.fixture
+def module_under_test():
+    import cdap.client
+
+    return cdap.client
+
+
+@pytest.fixture
+def object_under_test(module_under_test):
+    client = module_under_test.CDAPClient(credentials=None,
+                                          api_endpoint="http://localhost:11015")
+    client._transport = mock.create_autospec(requests)
+    return client
+
+
+def test_list_datasets_with_404(object_under_test):
+    mock_response = mock.create_autospec(requests.Response)
+    mock_response.status_code = 404
+    object_under_test._transport.request.return_value = mock_response
+    assert object_under_test.list_datasets() == []
+
+
+def test_list_streams_with_404(object_under_test):
+    mock_response = mock.create_autospec(requests.Response)
+    mock_response.status_code = 404
+    object_under_test._transport.request.return_value = mock_response
+    assert object_under_test.list_streams() == []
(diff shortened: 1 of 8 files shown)


tools/cdf-lineage-exporter/cdap/client.py
```diff
@@ -0,0 +1,143 @@
...
+            Optional. Namespace to use if not specified. Defaults to
+            ``'default'``.
+
+        credentials (Optional[google.auth.credentials.Credentials]):
+            ``google.auth`` credentials object to use in requests.
+    """
+
+    def __init__(self, api_endpoint: str, credentials=DEFAULT_CREDENTIALS, namespace="default"):
+        if credentials is DEFAULT_CREDENTIALS:
+            credentials, _ = google.auth.default(scopes=SCOPES)
```
jaketf: Yeah this is so much simpler thank you.
What output are you ignoring here?

tools/cdf-lineage-exporter/requirements.txt
```diff
@@ -0,0 +1,6 @@
+flask
+google-cloud-storage
+black
```
jaketf: curious why preference for black over yapf?
The style check for this repo checks against yapf.

Also we should probably remove black, isort, and pytest from this because there's no need for the runtime (cloud function) to use these static code / dev tools.

Thoughts on moving these to a `dev-requirements.txt` in the root dir of this example?

tools/cdf-lineage-exporter/tests/unit/test_client.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020 Google LLC
```
jaketf: remove empty test file or add unit test here.

tools/cdf-lineage-exporter/cdap/client.py
tswast: Second part of the tuple is `project_id`, which we don't need in this client.
https://google-auth.readthedocs.io/en/latest/reference/google.auth.html#google.auth.default

tools/cdf-lineage-exporter/requirements.txt
tswast: `black` is what Cloud DevRel has standardized on for all libraries and samples. Okay to switch back if professional services prefers `yapf`.

`dev-requirements.txt` is a good idea.

tools/cdf-lineage-exporter/tests/unit/test_client.py
tswast: Will do. There's a reason I opened this in draft mode. 😉

jaketf: yeah sorry about the premature review!

jaketf:
@tswast unfortunately and contrary to convention my experience of CDAP REST API is when you do a GET on an empty collection cdap returns 404 instead of empty list. That's why I added that decorator. If you add that decorator to all methods your integration tests will "pass" but we will not have fixed your issue. 
Book some time on my cal and let's figure out this permissions thing.

tools/cdf-lineage-exporter/requirements.txt
jaketf: Yeah the (currently failing) CI check uses `yapf --style=google`.
https://github.com/GoogleCloudPlatform/professional-services/blob/master/helpers/check_format.sh#L39

tswast:
Clearly there must be some difference between my version of yapf and the one Cloud Build is using. I've just pushed a commit that reformats using the exact same package versions via Docker.

```
docker run \
    --mount type=bind,source=/Users/swast/src/professional-services,target=/workspace \
    --workdir=/workspace \
    gcr.io/cloud-eng-council/make fmt
```

tswast:
@jaketf rebased and squashed. format check now passes, should be okay to review / merge

PR #427, opened 2020-03-18
fix: use yapf as a Python module in format/check_format

tswast:
This allows `make fmt` and `make test` to be runnable for Python files
in my local dev environment.


AdrienWalkowiak:
LGTM

PR #437, opened 2020-03-30
Automated dns for instances

vinson-rogers:
b/152636537

This example demonstrates how to use Cloud Asset Inventory to trigger cloud function to auto-populate Cloud DNS records for VM instance lifecycle events.

### DIFF
--- a/tools/automated-dns-for-instances/README.md
+++ b/tools/automated-dns-for-instances/README.md
@@ -1,28 +1,23 @@
-1. Overview
+# Automating Cloud DNS for VM Instances Tool
+
+## 1. Overview
 Using this document you will learn how to automate the creation and deletion of DNS ‘A’ records for compute instances on Google Cloud Platform.  We will use several GCP services to monitor virtual machine changes and act on specific events during their lifecycle. 
 
-
 The procedure involves the following steps:
-
 
 1. Creation of service account(s)
 2. Enable & Configure Cloud Asset Inventory (CAI) and Pub/Sub
 3. Deploy Cloud Function and Pub/Sub Trigger
 
-
 We will use the command line gcloud utility to configure and deploy all of the services. Installation for various environments can be found at https://cloud.google.com/sdk/install. Once installed you will need to authenticate & configure it:
 
-
 gcloud auth login
-	
 
 This will open a browser window and you will log in using your existing GCP user account. Once complete you can list the current gcloud configuration:
 
-
 gcloud config list
 
-# Example output:
-
+### Example output:
 
 [compute]
 region = us-central1
@@ -37,17 +32,12 @@ Your active configuration is: [default]
 
 Update the project, region and zone to reflect where you want to deploy the resources you will provision. We will use variables set in prior code blocks so stay in the same terminal for doing all of the work.
 
-
-2. Create Service Accounts
+## 2. Create Service Accounts
 You will need one or more service accounts for integrating the various services. You can create a single account and use it across Cloud Asset Inventory & Cloud Functions or create two separate ones. In this example we will create two separate accounts.
 
-
-2.1 SA Creation
+###2.1 SA Creation
 The following set of commands will create two service accounts and configure them with permissions to enable all of the functionality we need. Update your project name and change the names and descriptions of the service accounts if needed.
-
-
 The two service accounts will be owner/admin for their respective services in the project so it is critical to manage who can deploy code or use the accounts.
-
 
 PROJECT=[project_id_here]
 CAI_SA_NAME="cloud-asset-inventory-sa"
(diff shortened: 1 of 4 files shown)


tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+    if 'STAGING' in status and name != '' and ip != '':
+        print(f'Creating creation record set for VM: {name} with IP {ip}')
+        hostname = f'{name}.{DOMAIN}'
+        add_record_set = zone.resource_record_set(hostname, 'A', TTL, ip)
+        changes.add_record_set(add_record_set)
+
+    # execute changes for valid statuses
+    if any(x in status for x in valid_statuses):
+        changes.create()
+        #print(changes.status)
```
jaketf, line 81: remove commented out code

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+        while changes.status != 'done':
+            time.sleep(.1)
+            changes.reload()
+        #print('Changes status:', changes.status)
```
jaketf, line 85: remove

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+client = dns.Client(project=PROJECT_ID)
+zone = client.zone(ZONE, DOMAIN)
+
+def find_by_name(name):
+    records = zone.list_resource_record_sets()
+    for record in records:
+        if name in record.name:
+            rs = zone.resource_record_set(record.name, record.record_type, record.ttl, record.rrdatas)
+            return(rs)
+    return False
```
jaketf, line 33: nit: don't have different code paths with different return types when possible. Here  return `None` (which will be coerced  to `False` anyway)

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+def vmToDNS(event, context):
+    import base64, json
```
jaketf, line 36: move imports to top.

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+    changes=zone.changes()
+    valid_statuses = ['STAGING','DELETED']
+
+    if 'data' in event:
```
jaketf, line 40: nit: be more specific to dictionary use `dict.get` to raise exception if event is string (rather than trying to match substring data)
```suggestion
    if event.get('data'):
```

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+        data = base64.b64decode(event['data']).decode('utf-8')
+        data = json.loads(data)
+        # print(data)
```
jaketf, line 43: remove.

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+            return(rs)
+    return False
+
+def vmToDNS(event, context):
+    import base64, json
+    changes=zone.changes()
+    valid_statuses = ['STAGING','DELETED']
+
+    if 'data' in event:
+        data = base64.b64decode(event['data']).decode('utf-8')
```
jaketf, line 41: simplify to one line
```suggestion
        data = json.loads(base64.b64decode(event['data']).decode('utf-8'))
```

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+        data = json.loads(data)
+        # print(data)
+        if 'deleted' in data and data['deleted'] == True:
```
jaketf, line 44:
```suggestion
        if data.get('deleted'):
```

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+            status = 'DELETED'
+            # print(data['asset']['name'])
```
jaketf, line 46: remove

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+            match = re.search(r"/instances\/(.+)", data['asset']['name'])
+            # print(match[1])
```
jaketf, line 48: remove

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.cloud import dns
+import time
+import re
```
jaketf, line 17: alphabetize

tools/automated-dns-for-instances/main.py
```diff
@@ -0,0 +1,85 @@
...
+
+PROJECT_ID=<PROJECT_ID>
+ZONE=<ZONE_ID>
+DOMAIN=<DOMAIN> # should have a trailing dot
```
jaketf, line 21: Can we make this a "template file" that asserts that domain ends with dot?
e.g. make this
```
PROJECT_ID=${PROJECT_ID}
ZONE=${ZONE_ID}
DOMAIN=${DOMAIN}  # should have a trailing dot
```
and use something like [`enbsubst`](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html) in a small shell script. or wrap this in a small terraform module for deployment and use [`template_file`](https://www.terraform.io/docs/providers/template/d/file.html)

tools/automated-dns-for-instances/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+Your active configuration is: [default]
+```	
+
+Update the project, region and zone to reflect where you want to deploy the resources you will provision. We will use variables set in prior code blocks so stay in the same terminal for doing all of the work.
+
+## 2. Create Service Accounts
+You will need one or more service accounts for integrating the various services. You can create a single account and use it across Cloud Asset Inventory & Cloud Functions or create two separate ones. In this example we will create two separate accounts.
+
+### 2.1 SA Creation
```
jaketf, line 43: Would prefer to see Terraform for all these bits.
Less error prone / better user experience than gcloud commands.

jaketf: See @kardiff18's [similar asset](https://github.com/GoogleCloudPlatform/professional-services/tree/master/examples/ml-audio-content-profiling) for example

jaketf:
@vinson-rogers this is about to be closed by stalebot...

PR #461, opened 2020-04-28
Include eventid and eventtimestamp fields as pubsub attributes

prathapreddy123:
Pull request includes following changes:

1. Validate to ensure the schema is well formed json schema
2. Inspect schema for presence of event metadata fields such as eventid, eventtimestamp.
3. If event metadata fields are available include these fields as pubsub message attributes
4. Upgraded Beam version to 2.19 from 2.6


examples/dataflow-streaming-benchmark/README.md
```diff
@@ -39,6 +39,11 @@ instruct the data generator of what type of fake data to create in each field. S
 json-data-generator [docs](https://github.com/vincentrussell/json-data-generator) for more
 information on the faker functions.
 
+#### Message Attributes
+In case schema contains fields matching (case insensitive) names eventId, eventTimestamp then such fields
+will be added to attributes.  Attribute fields can be helpful in various scenarios like deduping messages, 
+inspecting message timestamps etc
```
ryanmcdowell, line 45: Cleaning up grammar... I would update it to something like the follow:

```
If the message schema contains fields matching (case-insensitive) the following names then such fields
will be added to the output Pub/Sub message attributes:
eventId, eventTimestamp

Attribute fields can be helpful in various scenarios like deduping messages, inspecting message timestamps, etc.
```

examples/dataflow-streaming-benchmark/README.md
```diff
@@ -71,6 +76,8 @@ Pub/Sub topic.
   "completed": false
 }
 ```
+Since schema includes eventId, eventTimestamp fields, message also contains these fields
+as attributes in addition to regular payload.
```
ryanmcdowell, line 80: Maybe update to:

```
Since the schema includes the reserved field names of `eventId` and `eventTimestamp`, the output Pub/Sub message will also contain these fields in the message attributes in addition to the regular payload.
```

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -180,8 +202,52 @@ public static PipelineResult run(Options options) {
       this.schemaLocation = schemaLocation;
     }
 
+    // Leave the scope as package private since its referred inside test methods for assertions
+    List<String> getAttributeFields() {
```
ryanmcdowell, line 206: Annotate with `@VisibleForTesting`

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -180,8 +202,52 @@ public static PipelineResult run(Options options) {
...
+      JsonObject jsonObject;
+
+      try {
+        // Generate sample message based on the provided schema.
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+          dataGenerator.generateTestDataJson(schema, byteArrayOutputStream);
+          payload = byteArrayOutputStream.toString();
+        }
+
+        LOG.info("Sample message: {}", payload);
```
ryanmcdowell, line 230: The message could be more descriptive as `Validating sample message: {}`

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -180,8 +202,52 @@ public static PipelineResult run(Options options) {
...
+
+      String payload;
+      JsonObject jsonObject;
+
+      try {
+        // Generate sample message based on the provided schema.
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
+          dataGenerator.generateTestDataJson(schema, byteArrayOutputStream);
+          payload = byteArrayOutputStream.toString();
+        }
```
ryanmcdowell, line 228: Can we encapsulate this code in another function given that it is repeated in the `processElement` method?

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -194,24 +260,31 @@ public void setup() throws IOException {
...
         payload = byteArrayOutputStream.toByteArray();
       }
 
+      if (this.includeAttributeValues) {
+        // Build a json Object, extract fields and include them in attributes
+        JsonObject jsonObject =
+            this.jsonParser.parse(new String(payload, StandardCharsets.UTF_8)).getAsJsonObject();
+        for (String field : attributeFields) {
+          attributes.put(field, jsonObject.get(field).getAsString());
+        }
```
ryanmcdowell, line 285: This step of deserializing the payload to pull out the attributes seems like it will create some overhead. Have you checked if the guidance in the README of ~2,500 QPS per core still remains valid in the case that we are publishing with attributes.

examples/dataflow-streaming-benchmark/src/test/java/com/google/cloud/pso/pipeline/StreamingBenchmarkTest.java
```diff
@@ -114,12 +157,12 @@ public void testMessageGeneratorInvalidSchema() throws IOException {
...
-      assertThat(message, is(notNullValue()));
-      assertThat(new String(message.getPayload()), is(equalTo(schema)));
-      return null;
-    });
+    //    PAssert.that(results).satisfies(input -> {
+    //      PubsubMessage message = input.iterator().next();
+    //      assertThat(message, is(notNullValue()));
+    //      assertThat(new String(message.getPayload()), is(equalTo(schema)));
+    //      return null;
+    //    });
```
ryanmcdowell, line 165: Why comment out the assertion?

examples/dataflow-streaming-benchmark/src/test/java/com/google/cloud/pso/pipeline/StreamingBenchmarkTest.java
```diff
@@ -128,19 +171,19 @@ public void testMessageGeneratorInvalidSchema() throws IOException {
...
-   * @param lines The lines to write.
+   * @param fileContent Content to write into the file.
    * @return The file written.
    * @throws IOException If an error occurs while creating or writing the file.
    */
-  private static ResourceId writeToFile(String filePath, List<String> lines) throws IOException {
+  private static ResourceId writeToFile(String filePath, String fileContent) throws IOException {
 
-    String fileContents = String.join(System.lineSeparator(), lines);
+    // String fileContents = String.join(System.lineSeparator(), lines);
```
ryanmcdowell, line 180: Uncomment or remove from the file.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
```diff
@@ -194,24 +260,31 @@ public void setup() throws IOException {
         }
 
         schema = byteArrayOutputStream.toString();
+        this.validateSchema(dataGenerator, schema);
```
ryanmcdowell, line 263: One of the benefits of allowing invalid schema is that it allows you to test how your pipeline handles malformed data. Do you think we should put this functionality behind a pipeline option or is the expectation that the user should cover that in their unit tests?

examples/dataflow-streaming-benchmark/README.md
prathapreddy123: Done

examples/dataflow-streaming-benchmark/README.md
prathapreddy123: Done

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
prathapreddy123: Done

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
prathapreddy123: Initially I thought the same but intentionally avoided (given that repeated code was only 2 lines) to reduce the number of function calls that needs to be made inside process element  during high volume stream of messages.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
prathapreddy123: Yes validated and performance overhead is very small.  Here are the benchmarks computed as average of 3 iterations:

1000 Messages - jsonOnly(293ms), json + DeSerialization(303ms)
2500 Messages - jsonOnly(494ms), json + DeSerialization(544ms)
5000 Messages - jsonOnly(796ms), json + DeSerialization(855ms)
10000 Messages - jsonOnly(1105ms), json + DeSerialization(1183ms)

examples/dataflow-streaming-benchmark/src/test/java/com/google/cloud/pso/pipeline/StreamingBenchmarkTest.java
prathapreddy123: With latest code the invalid schema would throw an exception and the same was asserted inside Test Annotation.  Also removing the commented code completely to avoid confusion.

examples/dataflow-streaming-benchmark/src/test/java/com/google/cloud/pso/pipeline/StreamingBenchmarkTest.java
prathapreddy123: Removed the line

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
ryanmcdowell: Sure but since this is running on the JVM the JIT compiler should inline it given it's frequency & size so the method call becomes close to free.

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
ryanmcdowell: LGTM. Thanks for validating!

examples/dataflow-streaming-benchmark/src/main/java/com/google/cloud/pso/pipeline/StreamingBenchmark.java
prathapreddy123: Since the message needs to be deserialized to extract the values for the special fields schema must be valid.  Added a pipelineoption called validateSchema that can be set to false to test pipeline with invalid schema.  However in such cases attributes will not be populated.

prathapreddy123:
@ryanmcdowell  - Thanks for reviewing.  Pushed an update incorporating the changes.

Issue #459, opened 2020-04-28
most assets have unit tests but CI infra doesn't run them


jaketf:
@morgante @AdrienWalkowiak @ryanmcdowell 
would love your thoughts on how we might tackle this.

## Issue
Contributors write tests that are never run and sometimes we merge changes that break tests.
we should have a part of the CI process for running tests in these assets.

## Background
Mostly the repo seems to be (mostly) standardized on
mvn test
python unittest
go test

### Test Discovery
Ideally we'd have a test discovery / running script in the CI tool similar to check_format.sh
However this is difficult to implement because each example / tool has different dependencies and sometimes use different build tools (npm, sbt, gradle, etc.)

### New Asset Bootstrapper
If we can align on approach we can write a "new tool / new example generator script that seeds a new directory w/ a template Make / Docker file, README, etc."

### Additional CI Check on PRs
For "every dir under tools / examples has Make / Docker / OWNERS / README

## More strict contribution guidelines requirements
I'd be interested in modifying the contribution guidelines so every new asset must have a Dockerfile / Makefile to run tests. This way discovery tool could just run `make test` in each directory. This would also make reviewers job easier to pull down and run tests (which I admittedly don't usually do).

Additionally, it feels like as we try to improve the quality / maintainability of this repo we should have an OWNERS file under each tool / example. It can be difficult the track down original authors to triage issues and it'd be nice to reach out to all owners for repo wide initiatives and say...

## Messaging
What's to best way to get this message out there (and do we agree with it): "Hey we all need to migrate out of exclusions list, add docker / make / owner files by end of 2020 and upgrade to python3 or your asset will be considered unmaintained and dropped from the repo"

morgante:
I think adding a Makefile to each directory is a good idea, we could easily update CI to run `make test` for each of those.

+1 to adding a new tool bootstrapper. I'd recommend using [cookiecutter](https://github.com/cookiecutter/cookiecutter) as a simple solution to add this.

In terms of rolling this out, I suggest we start by writing up a brief standards/contributing guide (or add it to go/pso-github-process). We can send out a call to action for PSO once that's written.

ryanmcdowell:
Agreed with the above proposal. 

Is there a reason why we wouldn't require each contribution to have it's own cloudbuild file? This is how I've seen this managed elsewhere where the parent build file can kickoff child builds (https://github.com/GoogleCloudPlatform/cloud-builders-community).

[...]

PR #469, opened 2020-05-04
test: add shellcheck to CI tests

tswast:
Towards https://github.com/GoogleCloudPlatform/professional-services/issues/459 and https://github.com/GoogleCloudPlatform/professional-services/issues/471


tswast:
I don't have permission to push images to `gcr.io/cloud-eng-council/shellcheck`, so someone else would have to build. (Or maybe I should add image builds to the central cloudbuild.yaml?)

morgante:
> Or maybe I should add image builds to the central cloudbuild.yaml?

That would be appreciated!

tswast:
> I'm not sure we should treat shellcheck differently. All other validations are centralized in check_format - is there a reason we couldn't use that approach for shellcheck?

I was thinking it'd be easier to determine failures when reviewing PRs if checks for different languages were separate (parallel) build steps.

Also, FYI: that `check_format.sh` script fails shellcheck hard. It'll take some refactoring / rewriting. I hope to split the other languages out too.

morgante:
Hmm, I think that could be a decent idea but we should probably still use a consolidated image and the possibility of running a single `make test` command to run all the tests (simpler DevEx IMO).

tswast:
> possibility of running a single make test command to run all the tests (simpler DevEx IMO).

I've spent way too many hours of my life trying to get a multi-language image to work for other mono-repos. It's much less overhead and much less flakey to maintain single-language containers.

morgante:
We already have it working for many other languages though (and have a similar image [here](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/tree/master/infra/build/developer-tools)). I'd like to avoid maintaining a dozen different images.

tswast:
Per the discussion on https://github.com/GoogleCloudPlatform/professional-services/issues/459, we're going to have a separate Cloud Build config for each project in this mono-repo. I don't see the benefits of trying to maintain a single image with

* Bash
* Python
* Go
* Java
* Scala
* Node.js
* Others?

It's clear that the current `make` image is insufficient for actually running tests in all these languages.

helpers/check_format_shell.py
```diff
@@ -0,0 +1,79 @@
...
+files.
+"""
+
+import pathlib
+import subprocess
+
+
+REPO_ROOT = pathlib.Path(__file__).parent.parent.absolute()
+
+ignored_filenames = set((
```
morgante, line 29: I don't know if we should have a separate exclusion list just for shellcheck?

Even if we do, I'd suggest:
1. Keeping this in a separate file, not inline in the script
2. Excluding directories/wildcards, not specific filenames.

morgante:
My understanding of #459 was that we would *add* additional Cloud Build configs for individual projects to run unit tests. This wouldn't be a replacement for the existing lint tests (which run on all projects automatically).

If we do want to start pushing images for each linter separately, I'd ask that you add a cloud build config for managing the images.

helpers/check_format_shell.py
tswast: > Keeping this in a separate file, not inline in the script

What difference does it make where it is?

> Excluding directories/wildcards, not specific filenames.

No.

If someone adds a new Bash script, I want this linter to run, regardless of if they have opted out other files in the past.

morgante: > What difference does it make where it is?

Config & code should be kept separate, that's a pretty fundamental software development idea. We shouldn't entangle config changes with logical changes.

Since it's a reasonably small number of files, I guess I'm fine with listing specific filenames.

tswast:
Build passes on my fork. https://console.cloud.google.com/cloud-build/builds/e2675fa3-d1c3-4322-81dc-3b265b9c38ef;step=0?project=swast-scratch

Will require rebuild of `make` image to pass on cloud-eng-council project

PR #476, opened 2020-05-08
Limiting flake to logical checks

prathapreddy123:
In this PR:

 Flake8 is limited to [logic checks](https://flake8.pycqa.org/en/latest/user/error-codes.html#error-violation-codes) as yapf is used to for formatting related checks

Closes #476

### DIFF
--- a/helpers/check_format.sh
+++ b/helpers/check_format.sh
@@ -45,7 +45,7 @@ validate_python() {
     # Initialize FILES_TO_LINT to empty string
     FILES_TO_LINT=""
 
-    (cd $FOLDER && flake8 --exclude=.git,__pycache__,.venv,venv --select=E9,F)
+    (cd $FOLDER && flake8 --exclude=.git,__pycache__,.venv,venv --select=E9,F,C)
     if [ $? -ne 0 ]
     then
        need_formatting $FOLDER
@@ -195,4 +195,3 @@ do
         echo "$FOLDER in exclusion list - SKIP  "
     fi
 done
-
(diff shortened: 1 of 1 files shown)


helpers/check_format.sh
```diff
@@ -45,7 +45,7 @@ validate_python() {
     # Initialize FILES_TO_LINT to empty string
     FILES_TO_LINT=""
 
-    (cd $FOLDER && flake8 --exclude=.git,__pycache__,.venv,venv --ignore=E501,W5)
+    (cd $FOLDER && flake8 --exclude=.git,__pycache__,.venv,venv --select=E9,F)
```
jaketf, line 48: nit: It would be valuable to also have [complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) checks provided by [mccabe plugin](https://github.com/PyCQA/mccabe) by default [uses C class of errors](https://flake8.pycqa.org/en/latest/glossary.html#term-error-code).
```suggestion
    (cd $FOLDER && flake8 --exclude=.git,__pycache__,.venv,venv --select=E9,F,C)
```

prathapreddy123: Included complexity check as well

jaketf:
@prathapreddy123 thanks for recognizing pain in the contributor experience for beam python on this repo and taking the initiative to work with our peers to find a solution!

prathapreddy123:
> @prathapreddy123 thanks for recognizing pain in the contributor experience for beam python on this repo and taking the initiative to work with our peers to find a solution!
:thumbsup:

PR #484, opened 2020-05-12
Remove GCP arch viz since it's moved to Forseti

ryanmcdowell:
The person who maintains GCP arch viz is no longer able to maintain it and the tool has been pulled into Forseti.

https://cloud.google.com/blog/products/identity-security/understand-gcp-organization-resource-hierarchies-with-forseti-visualizer


tools/gcp-arch-viz/README.md
```diff
@@ -1,42 +0,0 @@
-# GCP Architecture Visualizer
```
jaketf, line 1: IMO it would be good to leave a README here that points to the tool's new home.

ryanmcdowell: Would removing the sub-directory but modifying the root README to point to the following work or do we always want to keep the sub-directory?

https://github.com/forseti-security/forseti-visualizer

jaketf: that makes sense just put link in root readme.

ryanmcdowell: Updated

PR #485, opened 2020-05-16
set up directory for sql gravity

yzhvictor:
Set up the directory for bigquery sql gravity OSS projects

This summer, bigquery sql gravity team will welcome many OSS contributors to join and contribute to sql migration related tools. We'd like to set up a directory to host all the projects.


jaketf:
I'm open to these tool living in this repo but would a better home be https://github.com/GoogleCloudPlatform/bigquery-utils

In that repo it will be easier to add more specific CI tooling and will be more one stop shopping as we maintain UDFs to aid SQL migrations as well.

PR #492, opened 2020-05-21
Deployable Webhook App

dhercher:
A tool which deploys an unauthenticated webhook solution for loading data sent to an endpoint to BigQuery.

SDK or Webhook Data --> App Engine --> Pub/Sub --> Dataflow --> BigQuery

The design uses a simple App Engine design to receive data and forward it to Pub/Sub.
The PubSubCdcToBigQuery Flex template is used to transform data, manage new DDL on BigQuery tables, and finally load data into BigQuery.


tools/webhook-endpoint/README.md
```diff
@@ -0,0 +1,17 @@
+# Webhook App Engine Tool
+## Webhook App Overview
+The webhoook tool is intended to be an easy to deploy and easy to manage
+application.  You can use it to recieve and process any un-authenticated http(s)
+data.
```
jaketf, line 5: At first glance, un-auth'd data ingest seems like a great way to have a bad actor drive your cloud bill through the roof.
Can we please describe a motivating use case?

dhercher: This was a feature of alooma which we are trying to supply an open-source solution for.  Generally the most common use case here is a Mobile or Javascript SDK.  These are all open un-authenticated endpoints, so you could in theory go into someone's webpage or mobile app and abuse the web traffic.
This can also be used to replicate SaaS webhook data in realtime (ie. Stripe supplies webhooks like triggers which you can use to receive data in realtime)

This was always a nice way to allow different team to build their own integrations as well.  Often people wanted to build the reader and then just shoot data at Alooma's endpoint to manage creating and managing tables.  This option would be doable here w/ private IPs only for such use cases.

I think overall you'd have to lean on the DOS that's built into App Engine in this event.  The plan is to also use tokens which translate (base64) to the destination as a light extra layer.  Unknown or requests without tokens would be ignored.

tools/webhook-endpoint/.gitignore
```diff
@@ -0,0 +1,7 @@
+
```
jaketf, line 1: "webhook-endpoint" is a little non-descript of a directory name for a tool perhaps "webhook-data-ingest"?
Also you need an entry in the README at the root of the repo, please.

tools/webhook-endpoint/transforms/transform.py
```diff
@@ -0,0 +1,44 @@
...
+      if event:
+        print(event)
+  elif result:
+    print(result)
+
+if __name__ == '__main__':
+  # TODO: How do we handle the case where there are no messages
+  file_name = sys.argv[1]
+  with open(file_name, "r") as data_file:
+    for cnt, raw_data in enumerate(data_file):
```
jaketf, line 33: what determines how many elements will be enumerated here? beam bundling or something you control?

tools/webhook-endpoint/README.md
```diff
@@ -0,0 +1,17 @@
+# Webhook App Engine Tool
```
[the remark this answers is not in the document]
dhercher: The major driver is that the EPS doesn't match up well with Cloud Function or Cloud Run.  The scale would be hard to manage at a certain point but mostly they are both cost prohibitive for the use-case.
At 10,000EPS we are at 25billion per month
    - For App Engine this is ~$2000
    - Cloud Functions or Cloud Run is $10,000 base plus more for compute

App Engine also offers more flexibility here and IMO does a better job managing the app + ensuring portability

tools/webhook-endpoint/transforms/transform.py
dhercher: I tested out beams bundling and the batches were not configurable and not large enough to maximize the benefit.  The UDF has configurable windows and batch sizes that control this based on bytes.

tools/webhook-endpoint/deploy_dataflow.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Dataflow Config Vars
+NEW_UUID=$(head -n 50 /dev/urandom | tr -dc 'a-z0-9' | fold -w 6 | head -n 1)
+export DATAFLOW_JOB_NAME="webhook-job-${NEW_UUID}"
```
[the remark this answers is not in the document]
dhercher: It doesn't for the name though there is a custom ID.  This is our standard in Teleport docs and copied from there.
Long term I plan to add this deployment to terraform which will make all the management much easier.  I'm waiting on Flex to be offered there though

tools/webhook-endpoint/README.md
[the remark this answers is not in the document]
jaketf: Awesome, makes a ton of sense.
I'm still learning how to make these decisions between our server less offerings. 
TIL it's very important to crunch the numbers on cost for high request volumes!

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        publish_data(request_json, wait_for_ack=wait_for_ack)
+
+    return str(request_json)
+
+
+"""Publishes multiple messages to a Pub/Sub topic with an error handler."""
+def _get_callback(f, data):
+    def callback(f):
+        # TODO: implement logic here, this logic is okay but not great
```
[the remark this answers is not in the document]
dhercher: done

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return callback
+
+def publish_data(data, wait_for_ack=False):
+    """ Publish Pub/Sub Data """
+    if isinstance(data, dict):
+        data = json.dumps(data)
+
+    futures.update({data: None})
+    # When you publish a message, the client returns a future.
```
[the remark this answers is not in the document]
dhercher: done

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+    future = publisher.publish(
+        topic_path, data=data.encode("utf-8")  # data must be a bytestring.
+    )
+    futures[data] = future
+    # Publish failures shall be handled in the callback function.
+    future.add_done_callback(_get_callback(future, data))
+
+    # Wait for all the publish futures to resolve before exiting.
+    if wait_for_ack:
+        while futures:
```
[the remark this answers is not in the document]
dhercher: done

tools/webhook-endpoint/webhook_endpoint/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Webhook App Requirements
```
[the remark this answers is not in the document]
dhercher: done

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+    return str(request_json)
+
+
+"""Publishes multiple messages to a Pub/Sub topic with an error handler."""
+def _get_callback(f, data):
+    def callback(f):
+        # TODO: implement logic here, this logic is okay but not great
+        if f.exception():
+            print("Please handle {} for {}.".format(f.exception(), data))
```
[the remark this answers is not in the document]
dhercher: done, shifted to error

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+    elif isinstance(request_json, list):
+        for row in request_json:
+            publish_data(row, wait_for_ack=wait_for_ack)
+    else:
+        publish_data(request_json, wait_for_ack=wait_for_ack)
+
+    return str(request_json)
+
+
+"""Publishes multiple messages to a Pub/Sub topic with an error handler."""
```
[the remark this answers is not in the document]
dhercher: intended to move this whole block to separate module, done now

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+        <http://flask.pocoo.org/docs/1.0/api/#flask.Request>
+    Returns:
+        The response text, or any set of values that can be turned into a
+        Response object using `make_response`
+        <http://flask.pocoo.org/docs/1.0/api/#flask.Flask.make_response>.
+    """
+    # Send Request to PubSub
+    request_json = request.get_json(silent=True)
+    if request_json is None:
+        raise exceptions.WebhookException(consts.NO_DATA_MESSAGE, status_code=400)
```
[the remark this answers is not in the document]
dhercher: The idea of the app is to receive data.  If no data is sent IMO thats a bad request rather than requesting something thats not found.  So it made more sense for the no data in request to throw 400 over a 404

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+    # was GET or the credentials were invalid
+    return request.method
+
+# @app.errorhandler(exceptions.WebhookException)
+# def handle_invalid_usage(error):
+#     response = jsonify(error.to_dict())
+#     response.status_code = error.status_code
+#     return response
+
+def webhook_to_pubsub(request, wait_for_ack=True):
```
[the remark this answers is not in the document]
dhercher: actually this was copy/paste from our docs 😂 but I'll clean up the doc string and add a hint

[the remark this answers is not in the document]
dhercher: even if I believe type hints are non-pythonic and a clear Java engineer invasion of Python

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+    """HTTP Cloud Function.
```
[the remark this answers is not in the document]
dhercher: done

tools/webhook-endpoint/webhook_endpoint/main.py
```diff
@@ -0,0 +1,107 @@
...
+futures = dict()
+
+app = Flask(__name__)
+
+@app.route('/', methods=['POST', 'GET'])
+def receive_data():
+    if request.method == 'POST':
+        return webhook_to_pubsub(request, wait_for_ack=False)
+    # the code below is executed if the request method
+    # was GET or the credentials were invalid
```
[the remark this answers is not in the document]
dhercher: in future ill add support for get, for now im shifting to be only POST

tools/webhook-endpoint/README.md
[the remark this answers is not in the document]
dhercher: Yea, for for sure!  You'll also note that the webhook_to_pubsub function is actually an entry point for cloud functions though.  So it was designed to work for either and if there was a usecase under 2mil per month that would inside the free tier and a solid path

tools/webhook-endpoint/.gitignore
dhercher: Added to README
Renamed the directory to `webhook-ingestion-pipeline`

tools/webhook-endpoint/build.sh
```diff
@@ -0,0 +1,41 @@
...
+# NOTE: The Webhook App Deployment requires:
+# 	- The project to exist before deployment
+
+# Terraform Init and Apply
+cd terraform/ || exit
+echo "Apply Terraform Resources"
+terraform init
+terraform import google_app_engine_application.app "${PROJECT_ID}"
+terraform apply -auto-approve \
+	-var "project_id=${PROJECT_ID}" \
```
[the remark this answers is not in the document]
dhercher: done, the terraform naming looks weird to me but consistency wins out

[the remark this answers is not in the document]
dhercher: done to the extent I can.  Our linter doesn't accept the terraform lowercase env vars that I still need to use in the bash scripts.  Made a note to remove those when terraform offers dataflow flex support

tools/webhook-endpoint/webhook_endpoint/main.py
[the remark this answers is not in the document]
jaketf: I see I misread this get_json that htis was a GET request.
Agreed 400 looks good.

tools/webhook-endpoint/webhook_endpoint/main.py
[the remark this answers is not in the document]
jaketf: Cool Sounds like you should set return code to 405 Method Not Allowed for anything but POST then.

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+
+        return data
+
+    def _wait_for_ack(self, data, timeout=60):
+        """ Return when data is processed or raise timeout
+            
+            :param data: String data being processed
+            :param timeout: Int seconds to wait
+        """
+        for _ in range(timeout):
```
jaketf, line 76: nit:
because your loop does things other than wait (e.g. RPC calls to pubsub) it will actually wait significantly longer that 60s before timing out. 

This might be acceptable for your use case but goes against the spirit of the docstring.

A more precise pattern for this would be 
```python
from time import monotonic
....
poll_frequency =1
start_time = monotonic()
while monotonic() - start_time < timeout:
  if self.futures.get(data):
    time.sleep(min(0, timeout - (monotonic() - start_time), poll_frequency))    
```

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, project_id):
+        self.project_id = project_id
+        self.client = PublisherClient()
+        self.futures = dict()
+
+    def _get_callback(self, f, data):
+        def callback(f):
+            if f.exception():
+            	msg = "Please handle {} for {}.".format(f.exception(), data)
```
jaketf, line 36:
```suggestion
            	msg = f"Please handle {f.exeption()} for {data}."
```

tools/webhook-ingestion-pipeline/webhook_endpoint/main.py
```diff
@@ -0,0 +1,61 @@
...
+    return response
+
+def webhook_to_pubsub(request, wait_for_ack=True) -> str:
+    """ Return String response for HTTP Request Processing
+
+        :param request: (flask.Request) The request object.
+        <http://flask.pocoo.org/docs/1.0/api/#flask.Request>
+        :param wait_for_ack: Bool if we need to wait for Pub/Sub ack
+    """
+    request_json = request.get_json(silent=True)
```
jaketf, line 52: What happens if client post non-json payload will this raise and exception?
Should this be communicated back to the client as a 400 Bad Request?

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,82 @@
...
+            :param data: String data being processed
+            :param timeout: Int seconds to wait
+        """
+        for _ in range(timeout):
+            if self.futures.get(data):
+                time.sleep(1)
+            else:
+                return
+
+        raise WebhookException(consts.TIMEOUT_MESSAGE, status_code=408)
```
jaketf, line 82: Hmm 4xx error imply the issue was with the client.
408 specifically implies that the server timed out waiting for the client for request something which is not the case here.

In this case the server timed out waiting on dataflow to pubsub ACK. This sounds more like a 504 Gateway Timeout.

But do you really want to raise exception here? IIUC this means this webhook services has delivered the message to pubsub but it hasn't been ACK-ed by dataflow yet. 
This seems you've missed your SLO of 1 min delivery (because there's a backpressure dataflow is churning through or something) but the message has be committed to PubSub which will guarantee delivery.
Would you want the client to retry the request (in which case you'd publish a new pubsub message with new id and end up with duplicate data in BQ)?

It seems to me the client / end user doesn't need to know about this but you should log a warning that you're falling behind the SLO.

Context:
https://www.websitepulse.com/kb/4xx_http_status_codes
https://www.websitepulse.com/kb/5xx_http_status_codes

tools/webhook-endpoint/tests/system/send_data.py
```diff
@@ -0,0 +1,146 @@
...
+    """ Send Data at App Engine Endpoint
+        
+        :param batches: Number of batches to send (0 sends in infinite loop)
+        :param pool_size: Number of concurrent threads to use
+        :param request_size: Number of events per request to batch
+        :param batch_size: Number of events per request to batch
+        :param batch_sleep_secs: Seconds to sleep between batches
+    """
+    print("Test 4: Scale Testing")
+    pool = Pool(processes=pool_size)
```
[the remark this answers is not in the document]
dhercher: interesting read, I'll switch it over, from a light test it seems somewhat faster

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
dhercher: I agree on the 504.  Honestly it might make sense to remove this entirely.  For one thing is a code path thats not being used.

IMO there should be monitoring on the pubsub topic and dataflow over throwing a warning from app engine which is a service doing its job in that case.

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
dhercher: done

tools/webhook-ingestion-pipeline/tests/system/send_data.py
```diff
@@ -0,0 +1,130 @@
...
+
+    return app_domain
+
+def send_request(request_size, project_id=None):
+    # Send batches of data to speed up EPS
+    app_domain = _get_app_domain(project_id)
+    data = [get_row() for i in range(request_size)]
+    requests.post(app_domain, json=data)
+
+def test_scaling(project_id, batches=10, pool_size=10, request_size=200, batch_size=100, batch_sleep_secs=1):
```
jaketf, line 74: nit: don't name a function test* that that doesn't assert some behavior. 
this function does not assert anything about scaling behavior.
consider a name that is more specific to this functions purpose like: `generate_traffic`

tools/webhook-ingestion-pipeline/webhook_endpoint/consts.py
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Environment Variables
+PROJECT_ID = "GOOGLE_CLOUD_PROJECT"
+PUBSUB_TOPIC = "TOPIC"
+
+# Default Values
+MAX_CONTENT_SIZE = 1000000 * 5 # 5 megabytes
```
jaketf, line 20: nit avoid comments that could be improved var names.
```suggestion
MAX_CONTENT_MB = 1000000 * 5
```

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self, project_id):
+        self.project_id = project_id
+        self.client = PublisherClient(
+            client_config = {
+                "interfaces": {
+                    "google.pubsub.v1.Publisher": {
+                        "retry_params": {
+                            "messaging": {
+                                "total_timeout_millis": 650000,
```
jaketf, line 31: nit: don't use [magic numbers]
(https://en.wikipedia.org/wiki/Magic_number_(programming)#:~:text=In%20computer%20programming%2C%20the%20term,be%20replaced%20with%20named%20constants).
sanity check: this is 650 seconds did you mean 65 seconds? 

prefer: 
```python3
PUBSUB_TIMEOUT = (60 + 5) * 1000  # 1min plus 5s buffer
```

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,65 @@
...
+# limitations under the License.
+
+import json
+import logging
+
+from google.cloud.pubsub_v1 import PublisherClient
+
+
+class PubSubPublisher:
+
```
jaketf, line 22:
```suggestion
    _config = {
        "interfaces": {
             "google.pubsub.v1.Publisher": {
                 "retry_params": {
                     "messaging": {
                         "total_timeout_millis": 650000,
                  }
             }
          }
    }
```

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, project_id):
+        self.project_id = project_id
+        self.client = PublisherClient(
+            client_config = {
```
jaketf, line 26: nit: define this as a "private" static class member dict constant as it never changes.
Also clean up the formatting.

tools/webhook-ingestion-pipeline/tear_down.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Destroy Deployment
+cd terraform/ || exit
+bq rm -f --project_id="${PROJECT_ID}" webhook.webhook_data
```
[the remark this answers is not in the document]
dhercher: The table is one of the configurations of the Dataflow job, which is what controls the creation of the table and new columns.
tbh I use this more in testing to fully tear down the app, but in reality someone might want to keep the data even if they tear down the app, so I'll move this to the test for now

tools/webhook-ingestion-pipeline/webhook_endpoint/pubsub_publisher.py
dhercher: changed to const, using 10 min default but as a reference, and cleaned formatting

tools/webhook-ingestion-pipeline/tests/system/test_deploy_webhook.sh
```diff
@@ -20,8 +20,8 @@
 
 
 # Project Info for Testing Only
-export PROJECT_ID="my-testing-project"
-export PROJECT_NUMBER="my-testing-project-number"
+export PROJECT_ID="dhercher-webhook-21"
```
jaketf, line 23: don't leak you project id / number

tools/webhook-ingestion-pipeline/terraform/app_engine.tf
```diff
@@ -0,0 +1,118 @@
...
+}
+
+resource "google_storage_bucket_object" "app_engine_requirements" {
+  name   = "webhook_endpoint/requirements.txt"
+  source = "../webhook_endpoint/requirements.txt"
+  bucket = google_storage_bucket.webhook_gcs_stage.name
+}
+
+## App Engine:
+# TODO I'm removing this block since it fails on re-creation
```
morgante, line 54: +1 to removing the block and simply stating an app engine app is required.

tools/webhook-ingestion-pipeline/terraform/bigquery.tf
```diff
@@ -0,0 +1,43 @@
...
+  friendly_name               = var.bigquery_dataset
+  description                 = "Webhook Application Destination Dataset"
+  location                    = var.webhook_location
+  # default_table_expiration_ms = 3600000 # TODO why do we want a default?
+
+  labels = {
+    env = "default"
+  }
+
+  access {
```
morgante, line 27: Do they all need Owner?

tools/webhook-ingestion-pipeline/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+### Project Config ###
+variable "project_id" {
+  description = "Project ID for your GCP project"
+  default     = "my-project"
```
morgante, line 18: Don't provide a default for values which *must* be changed.

tools/webhook-ingestion-pipeline/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "project_num" {
+  description = "Project num for your GCP Project"
+  default     = "my-project-number"
```
morgante, line 23: Don't provide a default for values which *must* be changed.

tools/webhook-ingestion-pipeline/terraform/gcs.tf
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+resource "google_storage_bucket" "webhook_gcs_stage" {
+  project       = var.project_id
+  name          = var.gcs_bucket_name
```
morgante, line 18: I suggest either including the project ID or a random ID, as bucket names are globally unique.

tools/webhook-ingestion-pipeline/terraform/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+  description = "Pub/Sub Topic to Buffer App Engine to Dataflow Data"
+  default     = "webhook_topic"
+}
+variable "pubsub_subscription" {
+  description = "Pub/Sub Subscription to be used by Dataflow"
+  default     = "dataflow-subscription"
+}
+
+### Dataflow Config ###
+variable "dataflow_template" {
```
morgante, line 70: This variable isn't used anywhere?

tools/webhook-ingestion-pipeline/terraform/variables.tf
dhercher: removed these

tools/webhook-ingestion-pipeline/terraform/gcs.tf
dhercher: changed to prefix, use with project id as suffix

tools/webhook-ingestion-pipeline/terraform/bigquery.tf
dhercher: nope, removed all but the manager

PR #498, opened 2020-06-04
Patch/composer code lab

jaketf:
This lab was breaking because it was expecting python3 but default dataproc image was still python2. 
- add schema file that the codelab used to pick up from a public GCS bucket that has been deleted.
- fixed requirements.txt to acoomodate beam and airflow w/o errors (note this requires using  virtualenv instead of venv [AVRO-2776](https://issues.apache.org/jira/browse/AVRO-2776))
- specify a dataproc image in dataproc pyspark operator
- clean up use of airflow variables to use jinja template strings where possible.
- fix typo in tests
- add default python gitignore
- adds a cloudbuild.yaml to run the tests for these examples.

R: @tswast


examples/cloud-composer-examples/composer_http_post_example/ephemeral_dataproc_spark_dag.py
```diff
@@ -86,31 +78,33 @@
...
+        image_version='1.5-debian10',
         num_workers=2,
         storage_bucket=Variable.get('dataproc_bucket'),
         zone=Variable.get('gce_zone'))
 
     # Submit the PySpark job.
     submit_pyspark = DataProcPySparkOperator(
         task_id='run_dataproc_pyspark',
-        main=PYSPARK_JOB,
+        main='gs://' + Variable.get('gcs_bucket') +
```
jaketf, line 89: Note: frustratingly this field is not a template field (in default composer version of airflow 1.10.3)

jaketf: This retrieves the variable value at dag parse time rather than runtime.
This is sufficient for the codelab but not ideal.

mik-laj: Is this fixed in the master branch?

jaketf: @mik-laj if you mean is this fixed in airflow master, no.
[`DataprocSubmitPySparkJobOperator.template_fields`](
https://github.com/apache/airflow/blob/master/airflow/providers/google/cloud/operators/dataproc.py#L1259) does not contain main.

Opened a new issue https://github.com/apache/airflow/issues/9146

examples/cloud-composer-examples/composer_http_post_example/ephemeral_dataproc_spark_dag.py
```diff
@@ -132,17 +126,18 @@
...
 
     # If the spark job or BQ Load fails we rename the timestamped raw path to
     # a timestamped failed path.
-    move_failed_files = BashOperator(task_id='move_failed_files',
-                                     bash_command="gsutil mv gs://" + BUCKET +
-                                     "/{{ dag_run.conf['raw_path'] }}/ " +
-                                     "gs://" + BUCKET +
-                                     "/{{ dag_run.conf['failed_path'] }}/",
-                                     trigger_rule=TriggerRule.ONE_FAILED)
+    move_failed_files = BashOperator(
```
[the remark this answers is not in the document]
jaketf: IMO no.
What is the value in moving the files?

The point of this example is being able to have another process manually/directly trigger a DAG with REST API (e.g. once it is done dumping the raw files to a known prefix). The REST payload contains the prefix. If this other process has a the ability to control the prefix it writes to then there's no reason to re-use a prefix and leaving the files where they are make audit trail,  troubleshooting and certain replay scenarios easier.

Imagine you have a large body of data (transaction history) that you enhance with a look up table (side input that has latest user segmentation). It could be reasonable to say "I provide the currator of the look up table a REST API to re-run this DAG that runs this enhancement on the same data with their new look up values". This use case would be unecissarily complicated if this look up table gets updated rather frequently and we always moved the files to a success dir every time. This isn't really a backfill or rerun of a partition like we normally think of in airflow. To be blunt orchestration like [prefect](https://docs.prefect.io/) might better serve this use case but few people have those skills and GCP doesn't have a managed service for that 😃. 

A more interesting use of this pattern might be a dataflow streaming pipeline that performs some windowed aggregation before writing to target table. It could handle super late events by buffering them to files on GCS and kicking off an airflow DAG that's responsible for updating said aggregations in light of these super late events in the target table. It might be helpful to look directly at the buffered files where they are for troubleshooting and fire the same REST call to trigger the DAG again if necessary.

It's really dependent on use case if you want to move these files (imagine if there are other consumers of these input files that don't expect them to be moved).

examples/cloud-composer-examples/requirements.txt
```diff
@@ -1,3 +1,5 @@
-apache-airflow[gcp]==1.10.3
-apache-beam==2.15.0
+apache-airflow[gcp,crypto]==1.10.3
+apache-beam[gcp]<2.20.0
```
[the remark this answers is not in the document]
jaketf: Good point! done.

examples/cloud-composer-examples/composer_http_post_example/ephemeral_dataproc_spark_dag.py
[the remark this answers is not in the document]
prathapreddy123: I agree that it depends on the use case how the files should be treated post processing..

PR #507, opened 2020-06-16
Create a partition table for SLI/SLO.

Djabx:
For SLI/SLO report we only need last 28 days rarely more.
So it's a waste that the `report` partition is not partition.

This PR try to correct this case.

@ocervell


ocervell:
LGTM, could you please increase the version in the `setup.py` to the next minor version `1.1.0` ? Thank you

PR #513, opened 2020-07-13
Add solution diagram

manokhina:
Add solution diagram explaining what each of the step does, including the structure of tables created. 
Engineering Council bug ID: b/141161346


michaelwsherman:
SGTM

PR #531, opened 2020-08-03
Updated architecture diagram

onetwopunch:
Added updated architecture diagram to the Hashpipeline example


onetwopunch:
@mik-laj Thanks for the quick approval. Tests seem to be failing due to an error in tools/hive-bigquery which clearly have nothing to do with this. I rebased on upstream/master before pushing FWIW and the branch is up to date.

mik-laj:
I looked at this diagram in detail and it seems to me that it does not match the README.md file. Can you check if step 7 is correct?

onetwopunch:
@mik-laj Thanks for pointing that out, I missed a few things in the README as well. Updated now

PR #536, opened 2020-08-11
GCS Locking Tool 

nehanene15:
This tool uses the Google Cloud Storage [generation preconditions feature](https://cloud.google.com/storage/docs/generations-preconditions#_Preconditions) to create a consistent locking mechanism that can be used as a mutex. This can be used to do protected, globally serialized work within a Java [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) block.

We can use the fact that Google Cloud Storage is strongly consistent and that it provides object versioning (to prevent overwriting lock files) to create this distributed mutex.


tools/gcs-lock/README.md
```diff
@@ -0,0 +1,29 @@
+## GCS Lock
```
tswast, line 1: Do we plan on distributing this as a Maven package? If not, let's put this in the `examples/` directory.

nehanene15: Moved to `examples/`

examples/gcs-lock/README.md
```diff
@@ -0,0 +1,29 @@
+## GCS Lock
+
+Uses the [Google Cloud Storage generation preconditions feature](https://cloud.google.com/storage/docs/generations-preconditions#_Preconditions) 
+to create a consistent locking mechanism that can be used as a mutex. This can be used to do protected, globally 
+serialized work within a Java
+[try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) block. 
+
+See [this](https://github.com/marcacohen/gcslock) GitHub repository for a [Go](https://golang.org/) based GCS lock.
+
```
jaketf, line 9: One of the most popular interfaces for GCS from Java is the Hadoop Connector which has a cooperitive locking feature [blog](https://cloud.google.com/blog/products/data-analytics/new-release-of-cloud-storage-connector-for-hadoop-improving-performance-throughput-and-more) and [config docs](https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/CONFIGURATION.md#cooperative-locking-feature-configuration). This is a way of providing "directory" level semantics for mutating GCS resources.

Please mention this as an alternative for HDFS / GCS users as it will better suit this specific use case.

Please describe the motivating use case(s) for using this example and the choice for GCS.

I think using GCS as a mutex is an interesting example but can seem like a strange choice without some more context. At first glance I find myself asking "Why not use Firestore transactions or a locks DB cloud sql instance or resdis/memorystore?"
It seems to me that any of these systems could provide lower latency for claiming / releasing the lock than uploading GCS objects (gut feeling, not something I've benchmarked).

The answer may be "customer already had a GCS bucket and didn't want to manage / provision another technology for this use case" but we should explain that in a public facing README.

examples/gcs-lock/src/test/java/GcpLockTest.java
```diff
@@ -0,0 +1,105 @@
...
+    GcpLockFactory lockFactory = new GcpLockFactory(bucketName);
+
+    Assert.assertThrows(
+        RuntimeException.class,
+        () -> {
+          try (GcpLock gcpLock1 = lockFactory.createLock(LOCK_NAME, timeout, unit)) {
+            TimeUnit.SECONDS.sleep(3); // Do work
+            // Following requests for locking should be time out since resource is still locked
+            try (GcpLock gcpLock2 = lockFactory.createLock(LOCK_NAME, timeout, unit)) {
+              TimeUnit.SECONDS.sleep(3); // Do work
```
jaketf, line 60: No reason to sleep here.
In fact you could fail fast `self.fail("acquired second claim on unreleased lock")`

examples/gcs-lock/src/main/java/GcpLock.java
```diff
@@ -0,0 +1,85 @@
...
+     * backoff.
+     */
+    long startTime = System.nanoTime();
+    while (true) {
+      try {
+        BlobId blobId = BlobId.of(bucketName, blobName);
+        BlobInfo blobInfo = BlobInfo.newBuilder(blobId).setContentType("text/plain").build();
+        Storage.BlobTargetOption blobOption = Storage.BlobTargetOption.doesNotExist();
+        storage.create(blobInfo, blobOption);
+        System.out.println("Lock acquired.");
```
jaketf, line 57: remove this.
We shouldn't pollute std out.
This feels like a debug log at most.

examples/gcs-lock/src/test/java/GcpLockTest.java
```diff
@@ -0,0 +1,105 @@
...
+    /** Tests unlock/AutoClose. The method acquires one lock and releases it. */
+    GcpLockFactory lockFactory = new GcpLockFactory(bucketName);
+
+    try (GcpLock gcpLock1 = lockFactory.createLock(LOCK_NAME, timeout, unit)) {
+      TimeUnit.SECONDS.sleep(2); // Do work
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
+
+    Assert.assertEquals("Lock acquired.\n" + "Job finished. Unlocked.\n", bout.toString());
```
jaketf, line 79: Testing against print statements is not a good practice as it is very fragile and doesn't test behavior directly. For example if someone changes or removes your print statement this test will fail even thought the core functionality / behavior hasn't changed at all.

jaketf: Instead in your "Do work" section you can set a flag to true and assert that this flag is true.

examples/gcs-lock/src/test/java/GcpLockTest.java
```diff
@@ -0,0 +1,105 @@
...
+      e.printStackTrace();
+    }
+
+    try (GcpLock gcpLock1 = lockFactory.createLock(LOCK_NAME, timeout, unit)) {
+      TimeUnit.SECONDS.sleep(2); // Do work
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
+    Assert.assertEquals(
+        "Lock acquired.\n"
```
jaketf, line 99: same as above.

examples/gcs-lock/src/main/java/GcpLock.java
```diff
@@ -0,0 +1,85 @@
...
+        return;
+      } catch (StorageException e) {
+        long elapsedTime = System.nanoTime() - startTime;
+        // 412 Precondition Failed will indicate that the lock is already in use.
+        if (e.getCode() != 412) {
+          throw new RuntimeException("Error while locking: " + e);
+        } else if (elapsedTime >= timeout) {
+          throw new RuntimeException("Timed out while trying to acquire lock.");
+        } else {
+          System.out.println(
```
jaketf, line 67: Don't use print statements for logging.
Suggest using slf4j 
http://www.slf4j.org/

examples/gcs-lock/src/main/java/GcpLock.java
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void close() throws RuntimeException {
+    /** Method to release a lock implementing AutoClosable close(). */
+    try {
+      storage.delete(bucketName, "lock");
+      System.out.println("Job finished. Unlocked.");
```
jaketf, line 80: same as above don't use printing for logging.

examples/gcs-lock/src/main/java/GcpLock.java
jaketf: Reasoning:
The application using this mutex likely isn't running on a user's machine.
Logging utilities like slf4j have integrations with log aggregation systems like prometheus / stackdriver.

examples/gcs-lock/README.md
tswast: That's a great pointer to cooperative locking in the Hadoop Connector. Definitely something we'd want to encourage for those users. This example is a pessimistic lock.

We eventually want to supplement this sample to show it working from Dataflow, so it'll be a lot of connections to whatever mechanism we use for locking. Cloud SQL instances have a limited set of connections, so doesn't scale to the same degree as GCS. Firestore would be a good choice, but customer is already using GCS and not Firestore.

jaketf:
Per our chat, it seems [GCS suggests only updating objects once per second]( https://cloud.google.com/storage/docs/key-terms#immutability). 
This makes this solution feel not very scalable / general.
I'd like to pump the brakes on merging this PR as there may be better alternatives like [beam state api](https://beam.apache.org/blog/stateful-processing/).
We really need to describe the use case for when you'd want to use GCS to manage locks for this to be a useful example.

bharathkkb:
@jaketf I was listening to Neha's talk today and a file based lock pattern seems common in certain scenarios. A [similar approach](https://github.com/hashicorp/terraform/blob/6f4f6fbc72360cd8a7a408980b0acd19e8f7f5c1/backend/remote-state/gcs/client.go#L82) is used for Terraform state locking using gcs. IMHO this approach if documented with the caveats can solve general issues where we want to serialize operations.

jaketf:
I'm open to reopening this PR if we give some love to caveats docs. I think we should remove the suggestion for use within a dataflow pipeline because it seems like there are better tools in the beam framework for achieving the desired behavior. I agree there are more general things this could be useful for. I think it's most important that we have a good motivating example of it's use for including this java implementation in this repo. I don't think the world is without examples for the general concept using GCS for locking as there are several golang and python implementations maintained and released in other repos:
https://github.com/marcacohen/gcslock
https://github.com/thinkingmachines/gcs-mutex-lock

nehanene15:
I understand the skepticism over why a Java implementation is needed in this repo, especially since there are implementations in other repos. After feedback from the tech talk, I think it could be useful to include as a general pattern that PSO can provide to customers in the right scenario (e.g [parallelized computations in GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/jobs#managing_parallelism) or concurrent compute with serial IO to config files). 

I'd have to expand on the caveats, explain alternatives (Beam State API, GCS Connector), and detail what the "right" scenarios are though.

PR #541, opened 2020-08-12
Adding nlp-sentiment-analysis-pipeline repo

alouie510:
b/163434055

This repo is an example of a streaming Dataflow (Python) pipeline. It's purpose is similar to [kubeflow-pipelines-sentiment-analysis](https://github.com/GoogleCloudPlatform/professional-services/tree/master/examples/kubeflow-pipelines-sentiment-analysis).

The repo provides an example of a streaming pipeline that sends tweets from Twitter's API to Google Cloud Pub/Sub. Pub/Sub then sends the tweets to Dataflow, where it is analyzed using the Natural Language API. Sentiment results are stored in BigQuery.

NOTES:
* Please check if cloudbuild.yaml files are appropriate.
* No link put in README.md as it is under "Dataflow Python Examples"
* Passed formatter, linter, and Build


alouie510:
@googlebot I signed it!

jaketf:
we've update the default branch for this repo to "main"
https://sfconservancy.org/news/2020/jun/23/gitbranchname/
https://github.com/github/renaming
Please change target branch to main.

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+
+    Args:
+        bytestring: Byestring from Pub/Sub representing tweet JSON.
+
+    Returns:
+      Tuple of (`str`, `str`) representing:
+        ('<YYYY-MM-DD>', '<Full Tweet Text>').
+    """
+    json_value = json.loads(bytestring.decode('utf-8'))
+    time_value = datetime.fromtimestamp(int(json_value['Timestamp'])/1000.0).astimezone(timezone('US/Central')).strftime("%Y-%m-%d")
```
jaketf, line 58: should timezone be configurable?
Why default to US/Central not UTC?

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+            doc_index.add([time_value, 1])
+
+        window_start = int(window.start.to_utc_datetime().timestamp())
+        doc_index = doc_index.read().get(time_value, 0)
+
+        modified_key = "{time_value}+{window_start}+{doc_index}".format(time_value=time_value, window_start=window_start, doc_index=doc_index)
+        yield (modified_key, tweet_text)
+
+
+class NlpClient_Batched(beam.DoFn):
```
jaketf, line 128: nit: remove underscore use TitleCase for python classes.

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+
+        writer.close()
+
+        annotations = self.analyze_document(file_name)
+        for tweet in annotations.sentences:
+            yield (time_value, tweet.sentiment.score)
+
+        filesystems.FileSystems.delete([file_name])
+
+    @retry.Retry(predicate=retry.if_exception_type(exceptions.ResourceExhausted), initial=10.0, maximum=100.0, multiplier=2.0)
```
jaketf, line 174: Love that you're reusing this from google api core! really nice!

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+        (modified_key, text_iterable) = modified_tweet_tuple
+
+        time_value = modified_key[:modified_key.index("+")]
+
+        file_name = './' + modified_key + ".txt"
+        writer = filesystems.FileSystems.create(file_name)
+        punctuation = ['.', '?', '!']
+        for tweet in text_iterable:
+            tweet = tweet.replace('...', ' ').replace('\n', ' ')
+            for p in punctuation:
```
jaketf, line 162: You can do this more efficiently w/ compiled regex replace. 
https://docs.python.org/3/howto/regex.html#search-and-replace

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+    def process(self, modified_tweet_tuple):
+        if self.client is None:
+            self.client = language.LanguageServiceClient()
+
+        (modified_key, text_iterable) = modified_tweet_tuple
+
+        time_value = modified_key[:modified_key.index("+")]
+
+        file_name = './' + modified_key + ".txt"
+        writer = filesystems.FileSystems.create(file_name)
```
jaketf, line 158: There is no reason to use file system here as the scope of this "file" is just a single process element call and you ultimately just read the whole thing back into a string anyway.

It would be much faster (and more concise) to provide an in memory file-like object. Take a look at [StringIO](https://docs.python.org/3/library/io.html#io.StringIO) and use a [context manager](https://book.pythontips.com/en/latest/context_managers.html) for opening / closing.
Looks something like this:
```python3
import io
with open(io.StringIO) as tweet_doc:
    ...
    for tweet in test_iterable:
        tweet_doc.write(...)
    self.analyze_document(tweet_doc) # This should be refactored to take file-like object rather than file name.
```

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+            content=content,
+            type=enums.Document.Type.PLAIN_TEXT,
+            language=self.language
+        )
+
+        annotations = self.client.analyze_sentiment(document)
+        return annotations
+
+
+class NlpClient_ElementWise(beam.DoFn):
```
jaketf, line 189: Seem like you should have am abstract base class for NlpClient the DRY things up and have concrete base class for batched and elementwise.

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+            'window_start_utc': window_start,
+            'window_end_utc': window_end,
+            'sentiment_score_sum': data_tuple[0],
+            'num_tweets': data_tuple[1],
+            'average_sentiment_for_window': data_tuple[2]
+        }
+
+
+def run(argv=None, save_main_session=True):
+    parser = argparse.ArgumentParser()
```
jaketf, line 278: IMHO better to move arg parsing into it's own function.
Why? 
arg parsing has nothing to do w/ running the pipeline (e.g. happens only client side not on dataflow workers).
This is more clear if you have a function that parses args and a run command that just consumes those args.

This is just a suggestion.

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+    )
+
+    parser.add_argument(
+        '--bq_dataset',
+        dest='bq_dataset',
+        required=True,
+        help='Dataset ID for BigQuery table'
+    )
+
+    parser.add_argument(
```
jaketf, line 308: why are dataset / table separate args? should just take a `[project.]dataset.table` string

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/pipeline.py
```diff
@@ -0,0 +1,413 @@
...
+        '--fixed_window_size',
+        dest='window_size',
+        required=True,
+        help='Size of fixed windows in minutes'
+    )
+
+    parser.add_argument(
+        '--nlp_batching',
+        dest='nlp_batching',
+        default=False,
```
jaketf, line 325: seems like we should default to `True` for scaling while keeping api calls low.

examples/dataflow-python-examples/streaming-examples/nlp-sentiment-analysis-pipeline/src/beam/setup.py
```diff
@@ -0,0 +1,10 @@
+import setuptools
+
+setuptools.setup(
+        name='dataflow-twitter-sentiment-pipeline',
+        version='1.0',
+        author='Alex Louie',
+        author_email='my.email@google.com',
+        install_requires=[],
```
jaketf, line 8: seems like this requires google-cloud-language

jaketf (changes requested):

PR #543, opened 2020-08-13
Add STS Job Manager

danielbankhead:
Add STS Job Manager to `tools/`. STS Job Manager is a petabyte-scale bucket migration tool that manages large bucket-to-bucket transfers by utilizing parallel STS Jobs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ __pycache__/
 *$py.class
 dist/
 lib/
-!tools/sts-job-manager/lib
 *.egg-info/
 .cache
 pytestdebug.log
@@ -32,13 +31,11 @@ client_secrets.json
 .python-version
 .idea
 node_modules/
-.vscode
 *.code-workspace
 .envrc
 .coverage
 htmlcov/
 *.iml
-.venv
 
 # Terraform
 # See: https://github.com/github/gitignore/blob/master/Terraform.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ __pycache__/
 *$py.class
 dist/
 lib/
+!tools/sts-job-manager/lib
 *.egg-info/
 .cache
 pytestdebug.log
@@ -36,6 +37,7 @@ node_modules/
 .coverage
 htmlcov/
 *.iml
+venv
 
 # Terraform
 # See: https://github.com/github/gitignore/blob/master/Terraform.gitignore
@@ -48,4 +50,3 @@ override.tf.json
 *_override.tf.json
 .terraformrc
 terraform.rc
-
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ node_modules/
 .coverage
 htmlcov/
 *.iml
-venv
+.venv
 
 # Terraform
 # See: https://github.com/github/gitignore/blob/master/Terraform.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ client_secrets.json
 .python-version
 .idea
 node_modules/
+.vscode
 *.code-workspace
 .envrc
 .coverage
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+/tools/sts-job-manager/  danielbankhead@google.com
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-/tools/sts-job-manager/  danielbankhead@google.com
--- a/README.md
+++ b/README.md
@@ -1,114 +1,120 @@
 # Professional Services
+
 Common solutions and tools developed by Google Cloud's Professional Services team.
 
 [![Open in Cloud Shell](http://gstatic.com/cloudssh/images/open-btn.svg)](https://console.cloud.google.com/cloudshell/editor?cloudshell_git_repo=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fprofessional-services.git)
 
 ## Examples
+
 The examples folder contains example solutions across a variety of Google Cloud Platform products. Use these solutions as a reference for your own or extend them to fit your particular use case.
 
-
-* [Audio Content Profiling](examples/ml-audio-content-profiling) - A tool that builds a pipeline to scale the process of moderating audio files for inappropriate content using machine learning APIs.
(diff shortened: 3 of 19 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ __pycache__/
 *$py.class
```
boredabdel, line 17: Please add a *.json to avoid uploading service account json files

README.md
```diff
@@ -96,6 +96,7 @@ The tools folder contains ready-made utilities which can simpilfy Google Cloud P
 * [Site Verification Group Sync](tools/site-verification-group-sync) - A tool to provision "verified owner" permissions (to create GCS buckets with custom dns) based on membership of a Google Group.
 * [SLO Generator](tools/slo-generator/) - A Python package that automates computation of Service Level Objectives, Error Budgets and Burn Rates on GCP,  and export the computation results to available exporters (e.g: PubSub, BigQuery, Stackdriver Monitoring), using policies written in JSON format.
 * [Snowflake_to_BQ](tools/snowflake2bq/) - A shell script to transfer tables (schema & data) from Snowflake to BigQuery.
+* [STS Job Manager](tools/sts-job-manager/) - A petabyte-scale bucket migration tool
```
boredabdel, line 99: Please add a comment to explain what STS stands for

tools/sts-job-manager/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Assign the Appropriate Permissions
+
+#### STS Job Manager Permissions
+
+Ensure the [Storage Transfer API](https://console.developers.google.com/apis/api/storagetransfer.googleapis.com/overview) is enabled.
+
+Ensure the Storage Transfer Service service account has the [required permissions](https://cloud.google.com/storage-transfer/docs/configure-access).
+
+This tool requires an account with the following permissions:
```
boredabdel, line 33: an account or a Service Account ?

tools/sts-job-manager/README.md
```diff
@@ -0,0 +1,195 @@
...
+- `monitoring.metricDescriptors.create`
+- `monitoring.metricDescriptors.get`
+- `monitoring.metricDescriptors.list`
+- `monitoring.monitoredResourceDescriptors.get`
+- `monitoring.monitoredResourceDescriptors.list`
+- `monitoring.timeSeries.create`
+
+#### Pause All Jobs Permissions
+
+Requires an account with the following permissions:
```
boredabdel, line 56: Please be precise in this readme file with what do you mean by account ?

.gitignore
danielbankhead: Adding `*.json` would break many packages, namely JavaScript/Node.js, where `package.json`, `package-lock.json`, etc. are required.

Additionally, I'm not seeing a convention for service account file names, thus it would be difficult to determine whether a file is actually a service account file.

tools/sts-job-manager/README.md
danielbankhead: Updated and clarified

tools/sts-job-manager/README.md
danielbankhead: Updated and clarified in the above block

.gitignore
boredabdel: Yeah but you are not using Node.js in this project ? or is the PR missing some files ?

CODEOWNERS
```diff
@@ -0,0 +1 @@
+/tools/sts-job-manager/  danielbankhead@google.com
```
boredabdel, line 1: We do not support codeowners in the pso repo since you will not have admin access to the repo itself. Please remove this file

danielbankhead: Removed

.gitignore
danielbankhead: Not in this project, but there's about 504 existing JSON files in the repo utilized by JavaScript, Node.js, etc.

Additionally, as mentioned, I don't think it solves the root problem where the file may not be saved with `.json`.

boredabdel: Ahh i see what's happening here, you should not edit the .gitignore file in the root of the repo. Please add a .gitignore under tools/sts-job-manager and add the files you want to ignore there.

tools/sts-job-manager/lib/table_util.py
```diff
@@ -0,0 +1,90 @@
...
+from lib.services import Services
+
+
+def create_table(client: Services.bigquery, options: BigQueryOptions,
+                 table_name: str, schema: List[bigquery.SchemaField]):
+    """
+    Creates tables. Does not throw if it exists.
+    """
+
+    print("Creating `{}` table if it does not exists...".format(table_name))
```
boredabdel, line 34: Please make sure you are using a Logger library instead of print. If this is run inside a container i'm not sure you will proper structure logging. I'm just trying to be mindful to the use-case where people will want to run this on GCE or GKE. There is an example here https://github.com/GoogleCloudPlatform/professional-services/blob/master/tools/slo-generator/slo_generator/compute.py

.gitignore
danielbankhead: Added & migrated

tools/sts-job-manager/lib/table_util.py
danielbankhead: Good point - I'll switch to `logging`

boredabdel: Great, let me know when you change to logging and i will review it again. Thanks

danielbankhead: Updated - since enough code has changed I'll rerun a few e2e tests to ensure everything still works as expected. Then I'll follow-up with confirmation

.gitignore
jaketf: FWIW you can use pre-commit hooks to detect private keys rather than relying on filenaming convention and gitignore (note that this does not protect against someone renaming the key e.g. `mv my-key.json my-key.txt`).
All gcp SA keys have a private key in them ([docs](https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys)),
In other projects I use pre-commit to detect-private-keys https://pre-commit.com/hooks.html.
Perhaps it would be good for us to introduce a pre-commit in this repo though it may be difficult to make sure everything is compliant with all the checks we would want to add.

tools/sts-job-manager/lib/table_util.py
danielbankhead: Ok, we look gtg

.gitignore
boredabdel: Thanks @jaketf. I will look into adding this to the root repo

boredabdel (approved):
LGTM, @ryanmcdowell can you please or someone with admin privileges merge this

Thanks

Issue #533, opened 2020-08-10
Asset Inventory : google.api_core.exceptions.BadRequest: 400 Too many fields: 10090


jf-marquis-Adeo:
Hello
Dataflow pipeline is broken since 6th of august due to quota error: Too Many fields 10090!
```
{
    "insertId": "2411828167005042233:27398:0:130255",
    "jsonPayload": {
      "job": "2020-08-09_12_00_10-9423103018968181521",
      "logger": "root:batchworker.py:do_work",
      "thread": "69:140260486854400",
[...]

PR #544, opened 2020-08-18
Translate additionalProperties API properties into arrays.

bmenasha:
This changes the BigQuery schema and is not backwards compatible with prior
imports. Subsequent WRITE_APPEND import runs will fail with conflicting schema
errors. Users can archive existing datasets and start new ones if old data is
important.

Previously additionalProperties API properties were RECORD fields. As these are
user defined unstructured data they would just keep adding to the BigQuery table
schema as more and more user defined fields were added over time. This change
converts any additionalProperties API property into an array of name value pairs
where the value is the type of the additionalProperties property. This should
greatly reduce the number of columns in BigQuery tables taken up by user defined
labels or app engine version names.

This is another fix for Issue #533

### DIFF
--- a/tools/asset-inventory/README.md
+++ b/tools/asset-inventory/README.md
@@ -319,6 +319,75 @@ This repository contains some command line tools that let you run the export/imp
    ```
 
 
+## Input Schema Changes
+
+Sometimes the import schema will change. This can happen upstream when a backwards incompatible change is made to an API, such as a change in the datatype of a value or when the code in the pipeline changes the import format. This can cause imports with ```write_disposition=WRITE_APPEND``` to fail with an error simliar to this:
+
+  ```
+  "./asset_inventory/import_pipeline.py", line 422, in finish_bundle raise e
+  File "./asset_inventory/import_pipeline.py", line 419, in finish_bundle load_job.result()
+  File "/usr/local/lib/python3.7/site-packages/google/cloud/bigquery/job.py", line 733, in result return super(_AsyncJob, self).result(timeout=timeout)
+
+  File "/usr/local/lib/python3.7/site-packages/google/api_core/future/polling.py", line 127, in result raise self._exception google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table project-1:assets.run_googleapis_com_Revision. Field resource.data.spec.containers.livenessProbe.handler.exec.command has changed mode from NULLABLE to REPEATED [while running 'load_to_bigquery/load_to_bigquery']
+  ```
+
+To resume the import process there are three options.
+
+### 1. Delete the dataset and recreate it.
+
+The simplest option, this will discard all your data losing prior history but will let you continue the import. You can also try deleting the tables that fail to import if just a few of them.
+
+
+  ```
+  bq rm my_dataset_name
+  bq mk my_dataset_name
+  ```
+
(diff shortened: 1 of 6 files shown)


tools/asset-inventory/README.md
```diff
@@ -319,6 +319,75 @@ This repository contains some command line tools that let you run the export/imp
...
+  File "/usr/local/lib/python3.7/site-packages/google/cloud/bigquery/job.py", line 733, in result return super(_AsyncJob, self).result(timeout=timeout)
+
+  File "/usr/local/lib/python3.7/site-packages/google/api_core/future/polling.py", line 127, in result raise self._exception google.api_core.exceptions.BadRequest: 400 Provided Schema does not match Table project-1:assets.run_googleapis_com_Revision. Field resource.data.spec.containers.livenessProbe.handler.exec.command has changed mode from NULLABLE to REPEATED [while running 'load_to_bigquery/load_to_bigquery']
+  ```
+
+To resume the import process there are three options.
+
+### 1. Delete the dataset and recreate it.
+
+The simplist option, this will discard all your data losing prior history but will let you continue the import. You can also try deleting the tables that fail to import if just a few of them.
```
jaketf, line 338:
```suggestion
The simplest option, this will discard all your data losing prior history but will let you continue the import. You can also try deleting the tables that fail to import if just a few of them.
```

bmenasha:
Thanks Jake.  I updated the version  and added notes in the README.md for how to deal with this issue with some example queries for this specific change.  Because of the nature of the change, i don't think it's easy to create a view, but it should be straightforward to change specific queries and I provided an example of that. 
thanks

tools/asset-inventory/README.md
```diff
@@ -319,6 +319,75 @@ This repository contains some command line tools that let you run the export/imp
...
+  ```
+
+### 1. Change import configurations.
+
+to import to a new dataset, or set write_disposition=WRITE_EMPTY.
+Changing the dataset the pipeline imports to will create new tables or setting write_disposition to WRITE_EMPTY (which will delete existing data) will allow imports to resume. This is done by changing either the ```config.yaml``` if using the scheduled import process or the ```--parameters`` value in the gcloud command when invoking the template via gcloud.
+
+## Upgrading from version 1.0.0 to 2.0.0
+
+The 2.0.0 pipeline release unfortuantly changed the import schema to resovle [issue #533](https://github.com/GoogleCloudPlatform/professional-services/issues/533). Now some user specified and dynamic properties are represented as record arrays of key value pairs rather then just flat records. This was done to keep a more regular schema and prevent accumulation of columns from overflowing table limits. This change could require changes in how you query the data for example, previously to query the App Engine traffic split across two versions you would write:
```
jaketf, line 374:
```suggestion
The 2.0.0 pipeline release unfortuantely changed the import schema to resolve [issue #533](https://github.com/GoogleCloudPlatform/professional-services/issues/533). Now some user specified and dynamic properties are represented as record arrays of key value pairs rather then just flat records. This was done to keep a more regular schema and prevent accumulation of columns from overflowing table limits. This change could require changes in how you query the data for example, previously to query the App Engine traffic split across two versions you would write:
```

bmenasha: great thanks. done with latest amend.

tools/asset-inventory/README.md
bmenasha: done

PR #554, opened 2020-09-21
[slo-generator] Fix Prometheus window replacement to allow full expressions

ocervell:
Currently we cannot specify expressions that use `[window:resolution]` in Prometheus. This PR fixes that.


ocervell:
Moved to https://github.com/google/slo-generator

PR #560, opened 2020-09-24
Update to new billing export schema

kardiff18:
The billing export schema recently added in 3 new columns - updated the output query to comply with the new schema and fix any UNION ALL errors.


jaketf (approved):
You will be my best friend if you can set up some CI for the sql. 
E.g. can we set up these tables in the eng council project and `bq query --dryrun` them
This can wait for a later PR.

PR #569, opened 2020-10-06
add dataproc kerberos tf example

jaketf:
This adds an example terraform module for spinning up a kerberized data lake with cross realm trust.
R: @jhambleton 
CC: @roderickyao


examples/kerberized_data_lake/main.tf
```diff
@@ -0,0 +1,546 @@
...
+    "${var.metastore_cluster}_principal",
+    "${var.analytics_cluster}_principal",
+    "trust_${var.analytics_cluster}_${var.kdc_cluster}_principal",
+    "trust_${var.metastore_cluster}_${var.kdc_cluster}_principal",
+    "trust_${var.metastore_cluster}_${var.analytics_cluster}_principal",
+  ]
+}
+
+# This is a script to stage mock encrypted principals on GCS.
+resource "null_resource" "encrypted_principals" {
```
mik-laj, line 85: Should we add destroyer?

examples/kerberized_data_lake/main.tf
```diff
@@ -0,0 +1,546 @@
...
+      module.service_accounts.iam_emails["${var.metastore_cluster}-sa"],
+      module.service_accounts.iam_emails["${var.analytics_cluster}-sa"],
+    ])
+  ]
+  prevent_destroy = "false"
+}
+
+
+locals {
+  users = [
```
mik-laj, line 261: Should these be variables?

examples/kerberized_data_lake/main.tf
jaketf: done.

jhambleton (changes requested):
nice work jake, this looks great! reviewed with minor change requests.

examples/kerberized_data_lake/main.tf
```diff
@@ -0,0 +1,546 @@
...
+    "jake",
+    "jarek",
+    "kamil",
+    "daniel",
+    "core-data-svc",
+  ]
+  tenants = [for t in var.tenants : "${t}-svc"]
```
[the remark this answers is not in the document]
jaketf: Hmm I have a different thought. Requiring the user to use `-svc` naming convention is a leaky abstraction.
I think we should have separate vars for users (human) and tenant (non-humans) the `-svc` suffix is assumed for non-human principals by our init actions (implementation detail that should be abstracted away by this  module).

This TF module can enforce this with this line.
I've updated variables to have a separate users input.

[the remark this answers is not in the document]
jhambleton: agree with this update which makes it much cleaner.

jhambleton (approved):
lgtm

mik-laj (approved):

mik-laj:
🎉

license-templates/LICENSE.txt
```diff
@@ -0,0 +1,18 @@
+Copyright 2020 Google Inc.
```
mik-laj, line 1: This file is not used. Can you say more about it?

PR #571, opened 2020-10-10
Fix Python syntax error in bqpipeline.py

cclauss:
Output: https://github.com/cclauss/professional-services/actions

basestring, cmp(), raw_input(), and xrange were all removed from Python on 1/1/2020.

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.


jaketf (changes requested):
We use cloud build for CI in this repo and already run these static tests. 
You can take a look at this here: https://github.com/GoogleCloudPlatform/professional-services/blob/master/helpers/check_format.sh#L68
If you'd like to add codespell to the existing scripts I'd be happy to accept this PR.

Also, this CI was introduced fairly recently and clearly missed this bqpipeline asset as the original contributor opted out of CI via the exclusion list :( https://github.com/GoogleCloudPlatform/professional-services/blob/master/helpers/exclusion_list.txt#L6

The proper way to fix this would be to update bqpipeline to pass CI checks and remove it from the exclusion list.

cclauss:
Your Re-review please.

jaketf (approved):
LGTM thanks for the fixup!

PR #577, opened 2020-10-28
Ignore properties with no valid characters.

bmenasha:
- Drop resource properties that don't contain any BigQuery valid characters from
the schema. Previously such a property caused a 'string index out of range'
error during import.

- Upgrade to latest Beam SDK.


bmenasha:
i still want to submit this if someone can review.
thanks

jaketf (approved):
LGTM

PR #581, opened 2020-11-18
gce-usage-log: add filter on active projects

jaketf:
Fixes issue where unnecessarily searching deleted projects for instances which may cause errors.


tools/gce-usage-log/.gitignore
```diff
@@ -0,0 +1 @@
+target/
```
jaketf, line 1: only did this cuz when i pulled it down to run the tests I accidentally almost committed it :)

soellman (approved):
LGTM. Thanks for handling the PR!

PR #582, opened 2020-12-04
fix mapping between region specified and location for multi-regional …

jhambleton:
@jaketf - requesting review


examples/kerberized_data_lake/variables.tf
```diff
@@ -90,3 +90,16 @@ variable "users" {
...
+  type        = map(string)
+  default     = {
+    us            = "us"
+    northamerica  = "us"
+    southamerica  = "us"
+    europe        = "eu"
+    asia          = "asia"
+    australia     = "asia"
+  }
+}
```
jaketf, line 105:
```suggestion
}

```

jaketf (approved):
Thanks LGTM small suggestion.

examples/kerberized_data_lake/variables.tf
jhambleton: formatting updated

PR #589, opened 2021-01-05
Update gce-quota-sync.py

aos-aos:
Set project_id in the environment this script is running to avoid errors similar to https://stackoverflow.com/questions/47423772/how-to-set-project-id-to-avoid-warnings/65582521#65582521.


iht:
We are not using `master` as the main branch of this repo anymore. Please could you resubmit this PR and send it against the `main` branch?

Issue #595, opened 2021-01-16
Dataflow data generator: Call to undefined function data_gen.get_bq_schema_string() 


tommo245:
There is no get_bq_schema_string(). Changing it to data_gen.get_bq_schema() works. I will get on to forking/pull request now.

```
--- a/examples/dataflow-data-generator/data-generator-pipeline/data_generator_joinable_table.py
+++ b/examples/dataflow-data-generator/data-generator-pipeline/data_generator_joinable_table.py
@@ -144,7 +144,7 @@ def run(argv=None):
              # In this case we use the value passed in from the command line.
              data_args.output_bq_table,
              schema=None
-             if schema_inferred else data_gen.get_bq_schema_string(),
+             if schema_inferred else data_gen.get_bq_schema(),
```

PR #596, opened 2021-01-16
Change call to get_bq_schema_string()

tommo245:
To get_bq_schema() - that exists, the _string version does not. I've run this multiple times on Dataflow, it didn't run without changing this.

Fixes: https://github.com/GoogleCloudPlatform/professional-services/issues/595

Thanks


jaketf (approved):
Lgtm thanks

PR #599, opened 2021-01-25
598: Added explanation for gcs_output_prefix parameter in README

tommo245:
The parameter is just referred to as 'output_prefix'. Made it clear the parameter name is 'gcs_output_prefix'.


iht (approved):
I have checked that the `dest_name` for this option is `output_prefix`, but the user facing option name is actually `gcs_output_prefix`.

Thanks for fixing this.

LGTM

PR #600, opened 2021-01-26
[gmon] Support Cloud Monitoring API filters

ocervell:
Allows to use`--filters` / `--filter` to do API-side filtering (instead of response filtering on the client side) and thus adds other capabilities like API functions, or regex matches.

Example:
```
gmon metrics inspect istio.io/service/client/request_count -p <PROJECT_ID> --filter metric.labels.env=prod metric.labels.period="monitoring.regex.full_match(\".*d\")"
```


jaketf (approved):
LGTM

PR #605, opened 2021-01-30
Removing disrespectful terms: https://source.android.com/setup/contribute/respectful-code

waheedbrown:
Minor edit: removing [disrespectful terms](https://source.android.com/setup/contribute/respectful-code)


waheedbrown:
Bad pull request. Need to flatten old changes.

PR #607, opened 2021-01-30
gcs-bucket-mover: Fix extra retry caused by unhandled exception

woitaszek:
Primary fix:

* Add an explicit type conversion to prevent an unhandled exception in `_print_sts_counters` caused by comparing `str` and `int`. This exception was propagating up the call stack until interception by a `@retry` decorator, causing a second STS job to be created after the first STS job completed.

Ancillary changes:

* Add check confirming that source bucket properties could be retrieved at start of program
* Display bucket names in spinner messages at time of creation


eshen1991 (approved):

eshen1991 (approved):
Looks good.

iht (approved):
LGTM

PR #617, opened 2021-02-18
Add Secret Manager Helpers

vicenteg:
This Java library makes it easy to replace strings with Secret Manager payloads, especially in situations where files would ordinarily need to be stored with secrets embedded within them. The library implements Apache Commons Text interfaces to make replacing template strings with Secret Manager payloads simple.


tools/secret-manager-helper/.gitignore
```diff
@@ -0,0 +1,51 @@
+# Copyright 2017 Google Inc.
```
iht, line 1: Copyright 2021

tools/secret-manager-helper/src/main/java/com/google/cloud/secrethelper/SecretManagerStringSubstitutor.java
```diff
@@ -0,0 +1,46 @@
...
+                .put(
+                    "secretManager",
+                    StringLookupFactory.INSTANCE.interpolatorStringLookup(new SecretLookup(client)))
+                .put(
+                    "secretManagerToFilePath",
+                    StringLookupFactory.INSTANCE.interpolatorStringLookup(
+                        new SecretLookupToFile(client)))
+                .build(),
+            null,
+            true));
```
iht, line 44: This is confusing. I think it would be better to create each one of the arguments passed to `super` as local variables, and then use those local variables in the call to `super`.

iht (changes requested):
Please see the inline comments.

tools/secret-manager-helper/src/main/java/com/google/cloud/secrethelper/SecretManagerStringSubstitutor.java
vicenteg: Thanks for the comment. Agree it could be simplified, but: 

```
Call to 'super()' must be first statement in constructor body
```

I can simplify this by not using the builder:

```
    super(
        StringLookupFactory.INSTANCE.interpolatorStringLookup(ImmutableMap.of(
            "secretManager", StringLookupFactory.INSTANCE.interpolatorStringLookup(new SecretLookup(client)),
            "secretManagerToFilePath", StringLookupFactory.INSTANCE.interpolatorStringLookup(
                new SecretLookupToFile(client)))));
  }
```

Reasonable?

iht: Ah, right, apologies. Yes, that's clearer. Thanks.

vicenteg: Great, thanks. I pushed the change in 9ac582c5809a76aa4840dc599159ff49fb9662cc.

iht (approved):
LGTM

tools/secret-manager-helper/.gitignore
vicenteg: Thank you, fixed in bf9600a74b2db49037136c25ce4b9b9553467fdd.

PR #623, opened 2021-03-12
Update gmon to google-cloud-monitoring v2.x.x and above

ocervell:
Library stopped working because of the 1.x --> 2.x upgrade of the google-cloud-monitoring library, which breaks things. This bugfix version updates `gmon` to use the new interface offered in `google-cloud-monitoring`, and fixes the issues encountered when using the library.


rosmo (approved):
LGTM, seems there's no functional changes (mostly formatting + new library).

PR #639, opened 2021-04-14
Fix java formatting As it breaks Cloud Build

boredabdel:
google-java-format breaks when a Pull request contains Java code. That's because the list of files to format was passed via a variables with double quotes "".

google-java-format doesn't like that and complains cannot read file. It considered the input as a string which contains a long list of files.

Removing the quotes doesn't work, this is because the linter will complain about it (the next step in Cloud Build)

So the straightforward solution is to write the list of files to format temporary to a file and pass that list to the google-java-format via an @ directive which works.

Also updated some documentation


boredabdel:
@ryanmcdowell could you review this please for me :)

iht (approved):
LGTM

PR #642, opened 2021-04-16
remove quota-monitoring-alerting from the exclusion list

boredabdel:
remove quota-monitoring-alerting from the exclusion list


iht (approved):
LGTM

PR #649, opened 2021-05-04
Fix to support for non 'us' keys,  Add support to Global KMS keys and Key update

lcaggio:
- Add support for non 'us' regions
 - Keep disk device name
 - Add timestamp to keep disk name unique
 - Add support for Cloud KMS global keys

### DIFF
--- a/tools/gce-google-keys-to-cmek/README.md
+++ b/tools/gce-google-keys-to-cmek/README.md
@@ -1,21 +1,21 @@
-# Google-managed key to CMEK
+# Update GCE Disks Endription Key in use
 
-Convert disks attached to a GCE instance from Google-managed encryption keys
-to customer-managed encryption keys stored in Cloud KMS. This operation is
-performed in-place so the GCE VM instance maintains its current IP-address.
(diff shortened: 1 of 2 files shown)


tools/gce-google-keys-to-cmek/README.md
```diff
@@ -1,21 +1,21 @@
-# Google-managed key to CMEK
+# Update GCE Disks Endription Key in use
```
rosmo, line 1: Typo (encryption)?

tools/gce-google-keys-to-cmek/main.py
```diff
@@ -101,28 +109,32 @@ def migrate_instance_to_cmek(project, zone, instance, key_ring, key_name,
   for source_disk in disks:
     disk_regexp = r'^https:\/\/www\.googleapis\.com\/compute\/v1\/projects\/(.*?)\/zones\/(.*?)\/disks\/(.*?)$'
```
ludoo, line 110: I would move the pattern to a module-level constant (and maybe use split instead of a regexp anyway)

ludoo (approved):

PR #651, opened 2021-05-07
Fix main.py as it fails on VMs with multiple disks and regional keys

shashyajoshi:
1. The current code to get "region" from "zone" (`region = zone.rpartition("-")[2]`) produces wrong result
`>>> zone = "us-central1-a"`
 `>>> zone.rpartition("-")[2]`  # current code
` 'a'`                                           # wrong result
` >>> zone.rpartition("-")[0]` # proposed change
 `'us-central1'`                         # expected result

2. `key_name` variable assignment is inside the `for` loop. It fails on the second iteration (if there are more than one disks), as the `key_name` keeps getting added to its previous value. This is fixed by moving that line outside of the `for` loop, next to zone and region assignments

3. Line # 294 has an error with "end" which is not expected there with only one %s for "operation".

### DIFF
--- a/tools/gce-google-keys-to-cmek/main.py
+++ b/tools/gce-google-keys-to-cmek/main.py
@@ -99,7 +99,10 @@ def main():
 def migrate_instance_to_cmek(project, zone, instance, key_ring, key_name,
                              key_version, key_global, destructive):
   start = time.time()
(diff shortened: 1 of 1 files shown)


boredabdel (approved):
LGTM

PR #687, opened 2021-08-23
Bring example to Beam 2.31

iht:
Update Beam version to 2.31 (latest available at this moment), and fix a silly mistake with an unused variable in a test (duplicated code in the original).

Also, add some more hidden directories to .gitignore (for VS code users).

Tests are still passing in local.


elibaba (approved):
Looks good to me

tims (dismissed):

tims:
Step #0 - "lint": Some files need to be formatted in ./examples/bigtable-change-key - FAIL
Step #0 - "lint": ./examples/bigtable-change-key/src/test/java/com/google/cloud/pso/dofns/UpdateKeyTest.java
Step #0 - "lint": make: *** [Makefile:28: test] Error 1

Looks like it needs google style reapplied.

tims (approved):

PR #689, opened 2021-08-28
Remove f string for strings with no params

iht:
This fixes the build, which is failing currently just because of this.


boredabdel (approved):
LGTM

PR #690, opened 2021-09-07
Update google-cloud-support-slackbot README.md file to reduce risk of cases being filed to the Support team

TheLanceLord:
Description: The google-cloud-support-slackbot tool is not supported by the Google Cloud Support team, so I've added a line asking teams that use the application to file their issues on GitHub or to contact their account team with the aim of reducing the amount of support cases being opened for this.


iht (approved):
LGTM

Issue #601, opened 2021-01-26
e2e-home-appliance-status-monitoring/ Update to 2021 direly needed


btshow:
I tried this very interesting demo.
However, several parts of the code need to be updated/bug fixed to enable running it in year 2021. At least these parts: 

## 1. Tensorflow:
In the tensorflow code I needed to replace tf.contrib.estimator.stop_if_no_increase_hook with tf.estimator.experimental.stop_if_no_increase_hook) 

## 2. IoT Key Generation:
Key Generation for IoT (the following is no longer working:
"ssh-keygen -t rsa -b 4096 -f ./rs256.key
openssl rsa -in ./rs256.key -pubout -outform PEM -out ./rs256.pub")
I replaced it with and it seemingly worked:
openssl genpkey -algorithm RSA -out rsa_private.pem -pkeyopt rsa_keygen_bits:2048 
openssl rsa -in rsa_private.pem -pubout -out rsa_public.pem

## 3.AppEngine code:
There are authentification issues with the appengine (i still have to figure out how to solve these issues, but the app engine log says: 
"WARNING:googleapiclient.discovery_cache:file_cache is unavailable when using oauth2client >= 4.0.0 or google-auth" 
and 
"Traceback (most recent call last):
File "/env/lib/python3.7/site-packages/googleapiclient/discovery_cache/file_cache.py", line 33, in <module>
from oauth2client.contrib.locked_file import LockedFile
ModuleNotFoundError: No module named 'oauth2client'"

A code update or any help especially on this last app engine errors would be highly appreciated. Thank you.

JoseGuilhermeDoZe:
gs://gcp_blog/e2e_demo/ or

gs://gcp_blog/

is not accessible. It is not possible to understand the example provided without the files contained in this folder.
Please, you have any suggestions?

iht:
@lerrytang Maybe you can comment on this

lerrytang:
I've shared the data with @iht who will transfer that into a publicly accessible storage.
Thanks, Israel!

JoseGuilhermeDoZe:
@lerrytang @iht 
Thank you so much mr iht and mr lerrytang. Please is it possible to make the files available? I would very much like to study this specific example. My native language is not English, I apologize for errors

JoseGuilhermeDoZe:
@iht , @boredabdel , @lerrytang então? por favor, please, help, socorro .....

iht:
[...]

PR #711, opened 2021-10-29
Copy data from bucket that is not public anymore (and add Git LFS)

iht:
This copies some data from a bucket that is not available anymore, fixes some links (all except notebooks) to that data, and adds Git LFS configuration for the data files (they are not huge, the largest is 28 MB compressed, but it still makes operating with the repo slower unless LFS is used).

Partially addresses #601


boredabdel (approved):
LGTM

JoseGuilhermeDoZe:
@boredabdel @iht   Please where is the file https://storage.googleapis.com/gcp_blog/e2e_demo/model.tar  ??

PR #751, opened 2022-01-24
Service Account Provider - Workload Identity Federation workaround for Gitlab CI

cgrotz:
This is the source code for the Service Account Provider, a workaround for using GitLab CI together with Google Cloud APIs, to workaround some gaps in Workload identity federations OpenID Connect support.


helpers/exclusion_list.txt
```diff
@@ -23,6 +23,7 @@
 ./tools/ml-dataprep
 ./tools/ml-auto-eda
 ./tools/netblock-monitor
+./tools/service-account-provider/example_gitlab_pipeline
```
boredabdel, line 26: Why did you add this folder to the exclusion list ? Is the linting not passing ?

tools/service-account-provider/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+WORKDIR /go/src/app
+ADD . /go/src/app
+
+RUN go build -o /go/bin/app main.go
+
+FROM gcr.io/distroless/base-debian10
+COPY --from=build /etc/ssl/certs /etc/ssl/certs
+COPY --from=build /go/bin/app /
+
+CMD ["/app"]
```
boredabdel, line 26: Please use a multi-stage docker file ? https://docs.docker.com/develop/develop-images/multistage-build/

Essentially build using the golang image but the final image should not contain the compiler, you can use Alpine or Debian

tools/service-account-provider/README.md
```diff
@@ -0,0 +1,35 @@
+# Service Account Provider (SAPRO)
```
boredabdel, line 1: The Readme is missing instructions on how to setup and use the tool

boredabdel (changes requested):
Few comments

helpers/exclusion_list.txt
cgrotz: There are a few environment variables that are not set within the shell script (they are set from within the GitLab pipeline), and the linter was complaining about them. Since the script is just an example for how to use the tool, I added the exclusion.
Although it would be great if you have a suggestion how I could handle that differently.

tools/service-account-provider/Dockerfile
cgrotz: I am sorry, I thought I am using a multi-stage. Can you take another look, I am not sure what I am missing.

tools/service-account-provider/README.md
cgrotz: I added some instructions. Thanks for pointing this out.

boredabdel (approved):
LGTM

PR #753, opened 2022-01-26
Update BigQuery code to only return latest row

kimjulia1117:
Updated BigQuery to only return the latest row for each metric


boredabdel (approved):
LGTM

PR #757, opened 2022-02-03
pubsub2inbox: added MS Graph API mail transport + few minor fixes

mikouaj:
## New mail transport

This PR adds new mail transport for sending messages via [MS Graph sendMail API](https://docs.microsoft.com/en-us/graph/api/user-sendmail?view=graph-rest-1.0&tabs=http).
* the current implementation does not support attachments
* if both `html_body` and `text_body` are be defined, then `html_body` is used
* [Microsoft Authentication Library (MSAL) for Python]() with `ConfidentialClientApplication` are used to obtain oAuth2 `access_token` for MS Graph API authentication

## Minor fixes
* Docs: fixed `SERVICE_ACCOUNT` env variable referenced in README file to set service account for Cloud Function
* TF: added `bucket_location` variable for Cloud Function bucket location. Referenced it in `location` attribute in `google_storage_bucket.function-bucket`. The location attribute is required of Cloud Storage bucket
* TF: added resource to enable Secrets Manager API


rosmo:
LGTM, this is great! Even has tests!

boredabdel (approved):
LGTM

PR #766, opened 2022-02-17
DFDL Processor Example with Firestore and PubSub

carolinahdez:
# DFDL Processor Example
This module is a example how to process a binary using a DFDL definition with Daffodil libraries.
The DFDL definitions are stored in a Firestore database.

The application send a request with the binary to process to a pubsub topic.

The processor service subscribes to the topic, processes every message,
applies the definition and publishes the json result to a topic in pubsub.


boredabdel:
@carolinahdez Hi Carolina,

Thank you for the PR but it's pretty large (88 files). And it's modifyin a bunch of folders not only the tool you are submitting

Is this expected ? Generally speaking we prefer a PR to edit one tool at a time to make the review process easier.

Would it be possible to provide some context on the changes you are making here ? We would probably ask to break this down into smaller PR's

carolinahdez:
> @carolinahdez Hi Carolina,
> 
> Thank you for the PR but it's pretty large (88 files). And it's modifyin a bunch of folders not only the tool you are submitting

These PR only supposed to affect the example folder and add dfdl-firestore-pubsub-example folder only. I have tried several times to clean up but it keeps pulling those files. I am kind of lost with github. I do not use this tool.
> Is this expected ? Generally speaking we prefer a PR to edit one tool at a time to make the review process easier.
> 
No
> Would it be possible to provide some context on the changes you are making here ? We would probably ask to break this down into smaller PR's

PR #769, opened 2022-02-18
GCS to BQ using server-less services

bipinupd:
GCS to BQ Ingestion with server-less services (Cluod Function, PubSub and Serverless Spark)


examples/gcs-to-bq-serverless-services/Readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+        ```
+        bq_temp_bucket = <<GCS_TEMP_BUCKET>>
+        gcs_artifact_rep = <<GCS_ARTIFACT_REPO>>
+        dataset= <<DATASET_NAME>>
+        bq_table = <<TABLE_NAME>>
+        error_topic=<<ERROR_TOPIC>>
+        ```
+
+- **Step 5:** The cloud function is triggered one the object is copies to bucket.  The cloud function triggers the Servereless spark
```
iht, line 100: The cloud function is triggered **once** the object is **copied** to bucket.

examples/gcs-to-bq-serverless-services/gcs2bq-spark/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
iht, line 1: Missing copyright header

examples/gcs-to-bq-serverless-services/gcs2bq-spark/pom.xml
```diff
@@ -0,0 +1,208 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
iht, line 1: Missing copyright header

examples/gcs-to-bq-serverless-services/trigger-serverless-spark-fxn/requirements.txt
```diff
@@ -0,0 +1,3 @@
+requests
```
iht, line 1: Missing copyright header.

examples/gcs-to-bq-serverless-services/gcs2bq-spark/src/main/java/com/example/GCS2BQ.java
```diff
@@ -0,0 +1,167 @@
...
+      obj.put("projectId", projectId);
+      obj.put("inputFileLocation", inputFileLocation);
+      obj.put("bqDataset", bqDataset);
+      obj.put("bqTable", bqTable);
+      obj.put("deadLetterQueue", dlq);
+      publishMessage(obj.toString());
+      if (Objects.nonNull(spark)) {
+        spark.stop();
+      }
+    }
```
iht, line 117: This is too broad, it captures every single exception in the full pipeline. I think you should capture the exceptions that you expect to handle, and let the rest of exceptions be handled (or not) by Spark. The job would fail anyways.

iht (changes requested):
I have doubts about this example. This is triggering a full new job per file, instead of processing several files at once in the Spark job. Spark works by parallelizing in the input data. Here, if you receive 1k files, you will be launching 1k jobs. In terms of performance and cost, that's going to be much worse than launching a single job processing 1k files.

Also, the "failure publishing mechanism" should be done differently in my opinion.  The code captures all the exceptions in the full pipeline. That's too broad, and it may capture transient errors that could be recovered by Spark, by reprocessing the data.

Finally, I am missing unit testing for the pipeline at least (ideally, for the code of the CF too, but I think that's more difficult to test, as it only triggers the job).

boredabdel:
This PR is missing an entry to the root README.md file with a one line description of what the tool does.

examples/gcs-to-bq-serverless-services/Readme.md
bipinupd: > 

Thanks @iht for the review.

Agreed with your point, for the pipeline receives large number of small files, the example is not performant and cost effective. The example was meant for larger files (example daily extract from sources systems). I will make this explicit in Readme.
For exception handling, The example spark is simple (toy example), hence for any exception it is Publish the error to Pub/Sub and stops. I am also writing errors from cloud function to Pub/Sub. Not by default, but cloud functions are tried multiple times(and hence the spark), so yes multiple and transient error might go be written in Pub/Sub. I will like the idea of making sure the DLQ having only single entry failed jobs, so pipelines can be retired with less human intervention. Please share if you have some examples/patterns, I will incorporate them. Meanwhile I will also make it clear in Readme.

The solution is missing some unit tests, I will add it.

examples/gcs-to-bq-serverless-services/gcs2bq-spark/src/main/java/com/example/GCS2BQ.java
bipinupd: For exception handling, The example spark is simple (toy example), hence for any exception it is Publish the error to Pub/Sub and stops. I am also writing errors from cloud function to Pub/Sub. Not by default, but cloud functions are tried multiple times(and hence the spark), so yes multiple and transient error might go be written in Pub/Sub. I will like the idea of making sure the DLQ having only single entry failed jobs, so pipelines can be retired with less human intervention. Please share if you have some examples/patterns, I will incorporate them. Meanwhile I will also make it clear in Readme.

examples/gcs-to-bq-serverless-services/gcs2bq-spark/pom.xml
bipinupd: Added the copyright header.

examples/gcs-to-bq-serverless-services/gcs2bq-spark/dependency-reduced-pom.xml
bipinupd: remove the fie

iht (changes requested):
Thanks for the additional changes. The only part that is missing is the addition to the examples `README.md` file (a line, in alphabetical order, explaining this example), and checking the output of the Cloud Build failure.

iht (approved):
LGTM

PR #777, opened 2022-02-23
Custom Role Analyzer

dhananjaydalave:
While working with custom roles once created it is hard to find out if those roles are based out of any of the predefined roles.
In custom roles you will see a set of permissions as it is possible that many roles have common permissions.

This tool will provide useful insights with respect to custom roles at organization level as well as project level.
It will print results in file which contains following fields:
1. Custom Role - Name of the custom role.
2. Parent - Parent of the custom role. It can be an organization or project.
3. List of predefined roles - List of predefined roles by which this custom role is made of.
4. Additional permissions required - Any additional permission apart from predefined roles we need to build this custom role again.
5. Is Exact Match - If this custom role matches exactly to 1 or more predefined roles then it is set to True otherwise false.

It provides result in CSV or JSON format. Default is CSV.


examples/custom-roles-analyzer/docs/contributing.md
```diff
@@ -0,0 +1,28 @@
+# How to Contribute
```
boredabdel, line 1: This file exist in the root of the repo. It's not necessary to add it here.

examples/custom-roles-analyzer/docs/code-of-conduct.md
```diff
@@ -0,0 +1,93 @@
+# Code of Conduct
```
boredabdel, line 1: I think we should move this file to the root of the repo so it applies to the all the code not just this tool

boredabdel (commented):
Starting with few comments, Also please do the following:

Move the entire folder under tools/. This is a tool not an example
Have a look at the contributing.md and run the Linter and Formatter locally. It's failing in our CI

examples/custom-roles-analyzer/docs/contributing.md
dhananjaydalave: Got it. Removed this file.

examples/custom-roles-analyzer/docs/code-of-conduct.md
dhananjaydalave: Added this file at root level.

dhananjaydalave:
> 

Moved this to tools and fixed all formatting issues.

boredabdel (commented):
LGTM

boredabdel (approved):

PR #780, opened 2022-02-25
Add ability to specify folders and other filters when deploying the solution.

r-kyatham:
This change closes the last major feature gap(folder level scanning) between Python and Java versions of the solution. At the same time gives more configuration flexibility when compared to Java version.

Add 'Top 10 quota metrics with highest usage' chart to dashboard https://datastudio.google.com/c/u/0/reporting/50bdadac-9ea0-4dcd-bee2-f323c968186d/page/p_ou2cgupisc


boredabdel (approved):
LGTM

PR #784, opened 2022-03-05
Role bindings analysis feature

dhananjaydalave:
Added new feature:

Role bindings analysis
To find role binding at each resource level run this tool with --binding-analysis. This tool will write results in CSV format which include following fields:

Role - Any pre-defined or customer role.
Resource - Resource to which bindings are available.
Bindings - Bindings associated with resource and role.


boredabdel (approved):
LGTM

PR #790, opened 2022-03-14
Fix recursion bug in example/gcs-fixity-function

zefdelgadillo:
Fixes bug that would recursively run the function on all cloud storage files; not just the files within the targeted directory. This also adds a removal script.


boredabdel (approved):
LGTM

PR #793, opened 2022-03-16
DFDL Bigtable Pubsub Example

mijaz18:
## DFDL Bigtable Pubsub Example

An example to process binary string hexdump using DFDL and Daffodil libraries. The patch includes the code for DFDL processor, bigtable configuration and service, along with binary and DFDL examples.


examples/dfdl-bigtable-pubsub-example/src/main/java/com/example/dfdl/PubSubServer.java
```diff
@@ -0,0 +1,150 @@
...
+
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+      BasicAcknowledgeablePubsubMessage originalMessage =
+          message
+              .getHeaders()
+              .get(GcpPubSubHeaders.ORIGINAL_MESSAGE, BasicAcknowledgeablePubsubMessage.class);
+      originalMessage.ack();
+    };
```
iht, line 117: The ack of the original message will happen even if the lookup in Bigtable or the publish of the transformed message fail. Should not the ack happen only if the message can be successfully processed?

iht (changes requested):
I have a doubt about the ack of the incoming Pubsub messages. Please reply to the comment and/or make changes.

iht (approved):
LGTM

PR #801, opened 2022-03-31
Added creating private connection logic 

arunsingh2803:
Added more functionalities in the previously written automation script. Now user can also use private connection to create DataStream using oracle


boredabdel (approved):
LGTM

PR #804, opened 2022-04-06
A SPIFFE GCP Proxy

cgrotz:
This PR contains a small tool, that can be placed on a VM and behaves "like" a GCP Metadata server concerning tokens, but with SPIFFE and Workload Identity Federation as backend.


tools/spiffe-gcp-proxy/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+The proxy would run on a VM alongside a Spire agent. At the moment it's not possible to pass the identity of an application or process to the proxy, hence the proxies SPIFFE ID needs to be configured for impersonation in GCP IAM. In order to use the proxy from GCloud SDK or Google Cloud libraries, you should map `metadata.google.internal` to `127.0.0.1` in the VMs `/etc/hosts` file.
+
+After the proxy is setup you can just retrieve a GCP access token like you used to. This should work with proxy like gcloud, gsutil, ... or using curl with:
+````
+curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google"
+````
+
+In order to run the proxz will need to checkout the sources and use `go build main.go` to build the proxy. Afterwards you can run the binary but you will need to pass the required configuration via command line. See the full table of options below.
```
boredabdel, line 37: proxy ?

boredabdel (commented):
One typo to fix

tools/spiffe-gcp-proxy/README.md
cgrotz: Thanks, stupid german keyboard :-(

boredabdel (commented):
LGTM

boredabdel (approved):

Issue #808, opened 2022-04-15
[custom-roles-analyzer] Quota exceeded for quota metric 'SearchAllIamPolicies Requests' and limit 'SearchAllIamPolicies Requests per minute' of service 'cloudasset.googleapis.com'


jenson-gabriel:
We are running into a Quota limit error on the Cloud Asset API for SearchAllIamPolicies. The max quota for SearchAllIamPolicies is 400, which is the current value in the project. 

No other applications are using the Cloud Asset API in the current project.

If the query parameter is set to a specific group in AssetServiceUtil.java, then we did see the output CSV file generate successfully. The Custom Role Analyzer application should work once we get around this quota issue.

Error output:

```
Apr 11, 2022 10:03:02 PM com.google.cloud.pso.security.asset.AssetServiceUtil getAssetServiceClient
INFO: Get AssetServiceClient.
Apr 11, 2022 10:03:02 PM com.google.cloud.pso.security.asset.AssetServiceUtil createAssetServiceClient
INFO: Creating AssetServiceClient.
Apr 11, 2022 10:08:58 PM com.google.cloud.pso.security.CustomRoleAnalyzer main
SEVERE: Unable to process bindings for the roles.
com.google.api.gax.rpc.ResourceExhaustedException: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'SearchAllIamPolicies Requests' and limit 'SearchAllIamPolicies Requests per minute' of service 'cloudasset.googleapis.com' for consumer 'project_number:[REDACTED]'.
        at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:55)
        at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72)
        at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60)
        at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:97)
        at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:68)
        at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1133)
        at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:31)
[...]

PR #809, opened 2022-04-15
Fix for issue: 808

dhananjaydalave:
Fixing the issue:
[custom-roles-analyzer] Quota exceeded for quota metric 'SearchAllIamPolicies Requests' and limit 'SearchAllIamPolicies Requests per minute' of service 'cloudasset.googleapis.com' #808


boredabdel (approved):
LGTM

PR #814, opened 2022-04-26
mainframe connector phase 2 version

franklinWhaite:
This is a new version of the connector where all functionalities of the former are preserved.

The main change is that to reduce MIPs usage this version provides functionalities to use an intermediate GCE VM Instance to transcode COBOL binary, packed decimal and EBCDIC to partitioned ORC files in a streaming fashion and maximize computation offload from the mainframe.

Also, this project has been split into two projects 

- mainframe-util: It is added as a dependency to provide interfaces to commonly used services like Cloud Storage, BigQuery, and Cloud Logging. It also includes a custom SSLSocketFactory which forces use of TLS ciphers supported by IBM Hardware Crypto cards
- gszutil: provides gsutil and bq equivalent utilities for IBM mainframes running z/OS


boredabdel:
Hi @franklinWhaite.

Let's start with these two comments:

- Please add a license header to all source code files, including .sql, .sh and config files
- Please make sure the existing license headers in the files you changed are up to date (2022 instead of 2019)

Once done, ping me and i will review this one

Thank you

tools/bigquery-zos-mainframe-connector/README.md
```diff
@@ -1,326 +1,172 @@
...
-//QUERY DD DSN=<HLQ>.QUERY.FILENAME,DISP=SHR
+//QUERY DD DSN=<HLQ>.QUERY.FILENAME,DISP=SHR 
 //STDIN DD *
-bq query --project_id=myproject \
+PROJECT=my-long-project-name-1234
+bq query --project_id=$PROJECT \
   myproject:DATASET.TABLE \
   gs://bucket/tablename.orc/*
 /*
 ```
```
jasonmar, line 81: Recommend adding a note here stating that it is helpful to place commonly used project and bucket ids in the BQSH proc directly so that all JCL can reference them as environment variables. Based on the usage we've seen, different teams will likely have their own copy of the BQSH proc in their team-specific proclib, so the variables can be customized. This can also be used to seamlessly transition between prod and nonprod buckets/projects if environment-specific proclib and copy of BQSH are used.

tools/bigquery-zos-mainframe-connector/README.md
```diff
@@ -1,326 +1,172 @@
...
 
+### Enabling Debug Logging
+Debug logging can be enabled by editing the BQSH Procedure and setting the environment variable BQSH_ROOT_LOGGER=DEBUG \
+Debug logging can be disabled by commenting out the line or setting the variable to anything other than DEBUG.
 
-### Subcommand Usage
-
-#### gsutil
+## Using with Google Cloud Storage
```
jasonmar, line 87: recommend adding the gszutil command here and stating the difference between gsutil and gszutil as whether the grpc service is optional or not. (gszutil command is streamlined because it assumes the grpc server is used and this is the command expected to be used in prod for most cases)

tools/bigquery-zos-mainframe-connector/README.md
```diff
@@ -1,326 +1,172 @@
...
-  --debug_mode             Set logging level to debug. The default value is false.
-  --job_id <value>         The unique job ID to use for the request. If not specified in a job creation request, a job ID is generated. This flag applies only to commands that create jobs: cp, extract, load, and query.
-  --location <value>       A string corresponding to your region or multi-region location.
-  --project_id <value>     The project ID to use for requests. The default value is ''.
-  --synchronous_mode <value>
-                           If set to true, wait for the command to complete before returning, and use the job completion status as the error code. If set to false, the job is created, and successful completion status is used for the error code. The default value is true.
-  --sync <value>           If set to true, wait for the command to complete before returning, and use the job completion status as the error code. If set to false, the job is created, and successful completion status is used for the error code. The default value is true.
-  --stats_table <value>    tablespec of table to insert stats
+Check the [Using with BigQuery](./USING_WITH_BQ) guide.
```
jasonmar, line 93: the link doesn't seem to work when I view it on github

tools/bigquery-zos-mainframe-connector/README.md
```diff
@@ -1,326 +1,172 @@
...
-  -f, --force              delete without use interaction (always true)
-  destinationUri           Destination URI (gs://bucket/path)
-  --dataset_id <value>     The default dataset to use for requests. This flag is ignored when not applicable. You can set the value to [PROJECT_ID]:[DATASET] or [DATASET]. If [PROJECT_ID] is missing, the default project is used. You can override this setting by specifying the --project_id flag. The default value is ''.
-  --location <value>       A string corresponding to your region or multi-region location.
-  --project_id <value>     The project ID to use for requests. The default value is ''.
-  --allow_non_ascii        allow non ascii characters
-  --stats_table <value>    tablespec of table to insert stats
-  --max_error_pct <value>  job failure threshold for row decoding errors (default: 0.0
-```
+Check the [Using with Google Cloud Storage](./USING_WITH_GCS) guide.
```
jasonmar, line 89: the link doesn't seem to work when I view it on github

tools/bigquery-zos-mainframe-connector/USING_WITH_GCS.md
```diff
@@ -0,0 +1,89 @@
+# Using with Cloud Storage
+
```
jasonmar, line 2: recommend adding note about gszutil (grpc service required) vs gsutil and gszutil help text

jasonmar, line 2: also recommend a brief description of the functionality here and maybe some of the history.
"This project was created for loading Bigquery - the primary use case is copying a mainframe dataset directly to GCS, then invoking the associated gRPC service to transcode the mainframe dataset to ORC"
Users should pay attention to the amount of retention they'd like to have on the original mainframe datasets.
I think they're deleted after being successfully transcoded - please confirm that and note it here.

jasonmar, line 2: link to the readme for the gRPC service along with a minimal introduction.

jasonmar: I think it makes sense to put a description of how teams may deploy their own gRPC servers if they want dedicated transcoding capacity.

jasonmar: also helpful here is the number of parallel requests served by the gRPC service of a typical size. (maybe 8 cores, 16GB) and what happens when there are excess requests due to too many concurrent jobs and large datasets. The batch job side should time out the request at some point and users should probably expect the job to fail and be retried by their job scheduler, rather than the job running indefinitely waiting to be served.

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
```diff
@@ -0,0 +1,227 @@
+# Using with BigQuery
+
```
jasonmar, line 2: add notes about the job logging table and what information is recorded

jasonmar, line 2: add notes about the typical workflow being uploading a mainframe dataset, transcoding to ORC, then registering as an external table, then running a MERGE DML statement to load a daily incremental dataset into a target table

jasonmar, line 2: add note about using external table vs native table: if the dataset is unchanged after load, it makes sense to use a native table; if there are transformations, it is better to register as external then apply the transformations on the external table so that the data is only scanned once.

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
```diff
@@ -0,0 +1,227 @@
...
+## bq query
+
```
jasonmar, line 4: Add notes about the various ways queries can be provided - inline, separate dataset via DD, separate dataset via DSN.
Also add notes about how users may want to apply their own parameterization tokens to the query datasets separately from what is provided by the connector. This means that they would need a preceding job step to replace parameter tokens to produce the query dataset.

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
```diff
@@ -0,0 +1,227 @@
...
+  --synchronous_mode <value>
+                           If set to true, wait for the command to complete before returning, and use the job completion status as the error code. If set to false, the job is created, and successful completion status is used for the error code. The default value is true.
+  --sync <value>           If set to true, wait for the command to complete before returning, and use the job completion status as the error code. If set to false, the job is created, and successful completion status is used for the error code. The default value is true.
+  --stats_table <value>    tablespec of table to insert stats
+```
+
+
+
+## bq mk
+
```
jasonmar, line 155: add a note stating that this is most commonly used to create external tables or native tables that need partitioning and clustering to be set up

tools/bigquery-zos-mainframe-connector/gszutil/grecv/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+GCS is used as an artifact repository.
+Each version of grecv will have the following objects:
+- `mainframe-connector-version.jar` Application
+- `lib.tar` Application dependencies
+- `run.sh` Application launcher
+- `jre.tar` JVM
+- `pki.tar` TLS certificate and private key
+
+The `pkgUri` passed to grecv will be different for each environment.
```
jasonmar, line 47: pkgUri was added to launch a dedicated instance of this service for each job, but this ultimately wasn't used, I think we should remove this and all related documentation. The default deployment should be to long-running VM or container.

tools/bigquery-zos-mainframe-connector/gszutil/grecv/README.md
```diff
@@ -0,0 +1,90 @@
...
+The server accepts exactly the number of connections specified during the handshake.
+Before each data block, the client sends an integer indicating the size of the block.
+A value of less than 1 is sent to indicate end of stream.
+Upon receiving the end of stream message, the server sends the number of blocks it has 
+received and closes the connection.
+The client confirms the block count and closes the connection. 
+
+
+## Deployment
+
```
jasonmar, line 38: add some notes about sizing, and how separate organizations may want their own instance
also it may be helpful to note that currently the service is stateless so it could be placed behind a load balancer, but simply creating a larger instance should be able to serve most capacity needs even for a large enterprise, since the limiting factor is their upload bandwidth from the mainframe.
also it might be helpful to note that the service doesn't use any disk storage and logs directly to cloud logging.
I think systemd service logs are rotated automatically so there shouldn't be much issue running out of disk, and even if it did the service should continue to run with no problem.

tools/bigquery-zos-mainframe-connector/README.md
franklinWhaite: good idea, note added

tools/bigquery-zos-mainframe-connector/README.md
franklinWhaite: it should be working now

tools/bigquery-zos-mainframe-connector/README.md
franklinWhaite: it should be working now

tools/bigquery-zos-mainframe-connector/USING_WITH_GCS.md
franklinWhaite: documentation on gszutil has been added at the end of the file

franklinWhaite: details on this will be added to readme on gRPC

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
franklinWhaite: added

franklinWhaite: added

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
franklinWhaite: added

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
franklinWhaite: added

tools/bigquery-zos-mainframe-connector/gszutil/grecv/README.md
franklinWhaite: added

tools/bigquery-zos-mainframe-connector/USING_WITH_BQ.md
franklinWhaite: Added to the main readme since the table is supported by some bq and gsutil commands.

boredabdel (approved):
LGTM

PR #816, opened 2022-04-28
Add cloud-functions-act-as example

sshcherbakov:
Terraform project with GitHub Actions workflow that illustrates the way of Cloud Function invocation with workload identity and IAM permissions of the caller process, which is GitHub runner in this example.

The mechanism is effectively an "act as" mode of remote Cloud Function invocation which can be useful to reduce the Cloud Function service account IAM permissions to the least required privileges in cases which would otherwise require too wide IAM permissions to the resources of multi-tenant GCP organisation.

(No PSO Engineering Council bug id, since it is apparently on hold)


sshcherbakov:
@boredabdel I read the [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/professional-services/blob/main/CONTRIBUTING.md) 
and did Python linting as described in the referenced [styleguide](https://google.github.io/styleguide/).

It seems that shortcut was not enough :)

Rerun all steps from [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/professional-services/blob/main/CONTRIBUTING.md) and it seems that the Linting and Formatting CI step looks better now? Could you please let me know what are the remaining code style issues if any, since I can't seem to be able to see the build results.

boredabdel (approved):
LGTM

PR #820, opened 2022-05-04
Apigee bundle to convert JSON post to GET params

paulprogrammer:
Typical content mediation case, to transform a simple inbound JSON document to GET query parameters as an example of how to use Apigee to adapt different systems.


boredabdel (approved):
LGTM

PR #826, opened 2022-05-14
Adds an ansible playbook that enables Anthos Bare Metal users to easily change kubelet config of the cluster nodes

stackbaek:
Adds `update_kubelet_config` playbook that enables Anthos Bare Metal users to change kubelet config easily


boredabdel:
Hi,

Is changing the kubelet config something Anthos support and is recommended by our product team ? I feel like this is going to create more problems than solutions. Customer might feel it's something we recommend because there is a way to do it in the PSO repo. 

Do we have this documented somewhere in our docs ?  If so can you share the link ? Otherwise i don't really feel comfortable approving this

PR #828, opened 2022-05-20
Fix method of obtaining service account token

liggitt:
The .secrets field within service accounts is only for listing secrets
to be mounted into pods, not for extracting secrets for other uses.
There is no guarantee the first item in the list will be a token secret.

In 1.24+, this field is not populated by default.

Adapt instructions from https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-a-service-account-api-token to obtain a token

cc @zshihang


boredabdel (approved):
LGTM

PR #831, opened 2022-05-25
Projects recursion added to gcpohv-cli.py to fetch all projects instead of limited projects

dhruvjain92:
For GCP Organization Hierarchy Viewer, the projects were loaded initially using:

`projs = self._crm_v1.projects().list().execute()`

The problem with this approach was it didn't fetch all projects due to pagination. I have added a recursion function which uses the nextPageToken to get all projects and then manipulate them for the output.


boredabdel (approved):
LGTM

PR #835, opened 2022-06-06
Fix the documentation with only custom role analysis tool

dhananjaydalave:
Fix the documentation with only custom role analysis tool.


tools/custom-roles-analyzer/README.md
```diff
@@ -67,18 +72,17 @@ This will create an executable jar inside the `target` folder which will be used
...
 Run following command:
 ```
-analyze.sh --org <org_id> --format <format> --role-analysis --binding-analysis
+analyze.sh --org <org_id> --format <format> --role-analysis
 ```
 Parameters: \
 --org       : GCP Organization Id (mandatory parameter). \
 --format    : Result format (optional parameter; defaults to csv and supports json). \
 --role-analysis      : Run custom role analysis (Optional parameter; default to custom role analysis). \
---binding-analysis   : Run role binding analysis (Optional parameter; default to custom role analysis).
```
iht, line 76: I think this option is still in the tool. Why is it removed from here?

tools/custom-roles-analyzer/README.md
```diff
@@ -67,18 +72,17 @@ This will create an executable jar inside the `target` folder which will be used
...
 You can simply run as follows to get default format and custom role analysis:
 ```
 analyze.sh --org <org_id>
 ```
- 
+<!-- 
 You can run tool as follows to find the role bindings:
```
iht, line 86: Why is this commented out? Is it not valid anymore? I think those options still work.

tools/custom-roles-analyzer/README.md
```diff
@@ -33,24 +35,27 @@ It will print results in file which contains following fields:
 
 It provides results in CSV or JSON format. Default is CSV.
 
+<!--
```
iht, line 38: Why is this commented out?

iht (commented):
I don't understand the change in the documentation. The options that are removed are still in the code. Are not those options valid anymore?

tools/custom-roles-analyzer/README.md
dhananjaydalave: IAM team advised to write blog around role-analysis tool. The blog will point to this tool. So as to avoid confusion I am commenting out documentation for binding-analysis.

I am thinking to separate binding-analysis from role analysis. When I do that I will move the binding analysis code.

tools/custom-roles-analyzer/README.md
dhananjaydalave: I am planning to move binding-analysis to different tool. At this time just removing this from documentation and later will move the code.

tools/custom-roles-analyzer/README.md
dhananjaydalave: Those options will work until I move the code to different tool.

iht:
Then I think you should submit the changes in the documentation along with the changes in the code. The documentation should match the code at all moments.

PR #839, opened 2022-06-13
Bug fixes for the vm-migrator and troubleshooting info for vpc-flowlogs-toptalkers

lgandras:
Summary:

* Provide a workaround for GCP enforced throttling when creating VMs too fast
* Provide a workaround for when VPC flowlogs datasets are not available while creating the BQ view through terraform
* Fix the start instances command (it was shutting down the VMs)
* Release 3000 IPs by default (it was releasing 500 IPs by default).


rosmo (approved):
LGTM

PR #841, opened 2022-06-21
Disclaimer for java version and copyright updates

anuradha-bajpai-google:
Adding disclaimer for Java version continuation and updated copyright details with year


iht (approved):
LGTM. Thanks for keeping this asset updated!

PR #845, opened 2022-06-21
BigQuery Tink Toolkit

dmedora:
This PR includes a number of libraries and samples to help users make use of the Tink cryptography library in a way that is compatible with BigQuery's SQL-based field encryption, both deterministic and non-deterministic. 

Under the hood, BigQuery uses Tink for cryptography. We provide a library containing a couple of utility classes to help manage working with GCP KMS-wrapped Tink keysets outside of BigQuery. It also includes a Spark sample that shows how this could be used to encrypt/decrypt data (either on-prem or in GCP Dataproc) that resides in BigQuery. Additionally, it includes a script for generating these keysets.


iht (approved):
LGTM. Thanks for contributing to the repository!

PR #846, opened 2022-06-22
GCP SA to assume AWS ARN Role

danielmarzini:
A CLI utility to let a GCP Service account being able to assume AWS ARN Role when used from a VM/Container with the awscli.


danielmarzini:
@rosmo would you please review this one? Thanks!

PR #847, opened 2022-06-25
Add a Workload Identity Federation with Gitlab example

yuryninog:
Add a Workload Identity Federation with Gitlab example

This repository provides an example for creating a Workload Identity Federation (WIF) component that can be used for
authenticating to Google Cloud from a GitLab CI/CD job using a JSON Web Token (JWT) token.


yuryninog:
I have a bug in the definition of a variable.

PR #862, opened 2022-07-19
Implement Dataflow CsvIO example with Custom Template

damondouglas:
This PR provides two demonstrations.  The first is an example processing CSV files into [Schema](https://beam.apache.org/documentation/programming-guide/#what-is-a-schema)
aware PCollections using [Apache Beam](https://beam.apache.org/) where we expect
different headers.  Please consult examples/dataflow-csvio/README for an explanation of this problem and what this example solves.

The second is an example how to build and deploy the example as a custom Dataflow template using gradle in the context of terraform.  The community has expressed that documented Beam Java SDK examples only use maven.  Using Infrastructure-as-Code with terraform allows one to integrate the building of custom Dataflow templates within the workflow where one provisions dependent and related Google Cloud resources.


examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+    public abstract Builder setConfiguration(CSVIOReadConfiguration value);
+
+    public abstract SortContextualHeadersAndRows build();
+  }
+
+  /**
+   * The result of processing the raw {@link ContextualTextIO#read()} {@link Row} {@link
+   * PCollection} output into the corresponding CSV headers and rows.
+   */
+  static class Result implements POutput, PInput {
```
prathapreddy123, line 104: How about calling the class name as SortContextualResult to make it more explicit?

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+    SortFixedHeaderAndRows(CSVIOReadConfiguration configuration) {
+      this.configuration = configuration;
+    }
+
+    @Override
+    public Result expand(PCollection<Row> input) {
+      PCollectionView<Long> headerPositionView =
+          input
+              .getPipeline()
+              .apply(TAG_BASE + "/HeaderPositionView", Create.of(configuration.getHeaderPosition()))
```
prathapreddy123, line 169: apply method can be directly used on Pcollection instead of  Pcollection.getPipeline().apply

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+              .apply(View.asSingleton());
+      PCollectionTuple pct =
```
prathapreddy123, line 171: Better name would be headerAndRowsPCollectionTuple instead of pct

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+
+    SortMatchedHeaderAndRows(CSVIOReadConfiguration configuration) {
+      this.configuration = configuration;
+    }
+
+    @Override
+    public Result expand(PCollection<Row> input) {
+      PCollectionView<String> matchHeaderRegexView =
+          input
+              .getPipeline()
```
prathapreddy123, line 203: Same comments as earlier on usage of getPipeline as well as passing getHeaderMatchRegex directly instead of building PCollectionView

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+      PCollection<KV<String, Long>> resourceIdRecordOffsetKV =
+          input
+              .apply(
+                  TAG_BASE + "/" + IsNotEmptyFn.class.getSimpleName(),
+                  Filter.by(new IsNotEmptyFn()))
+              .apply(
+                  TAG_BASE + "/MapElements/" + RowKVFn.class.getSimpleName(),
+                  MapElements.via(new RowKVFn()));
+
+      PCollection<KV<String, Long>> minimumResourceIdRecordOffsetKV =
```
prathapreddy123, line 250: Need reevaluation of calculating recordnum during the cases when [withRecordNumMetadata()](https://beam.apache.org/releases/javadoc/2.38.0/org/apache/beam/sdk/io/contextualtextio/ContextualTextIO.Read.html#withRecordNumMetadata--) is not used and can be set as 1 (i.e default position)

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+
+    SortFixedPositionHeadersAndRowsFn(PCollectionView<Long> headerPositionView) {
+      this.headerPositionView = headerPositionView;
+    }
+
+    @ProcessElement
+    public void process(ProcessContext ctx) {
+      Long headerPosition = ctx.sideInput(this.headerPositionView);
+      Row input = ctx.element();
+      RecordWithMetadataHelper record = RecordWithMetadataHelper.of(input);
```
prathapreddy123, line 283: Since calls to RecordWithMetadataHelper happens per element this leads to creation & destruction of several objects causing excess GC overload.  Better to move RecordWithMetadataHelper as standalone class (in its own file) with static methods only and can be called passing row as Input

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+      Long recordNum = record.getRecordNum();
+      if (recordNum < headerPosition) {
+        return;
+      }
+      if (recordNum.equals(headerPosition)) {
```
prathapreddy123, line 288: Since Long is numeric, comparison can be recordNum == headerPosition

examples/dataflow-csvio/src/main/java/com/google/example/csvio/SortContextualHeadersAndRows.java
```diff
@@ -0,0 +1,427 @@
...
+   * Determines whether the CSV record {@link ContextualTextIO#read()} output is a non-empty line.
+   */
+  static class IsNotEmptyFn extends SimpleFunction<Row, Boolean> {
+
+    @Override
+    public Boolean apply(Row input) {
+      if (input == null) {
+        return false;
+      }
+      return !RecordWithMetadataHelper.of(input).getValue().isEmpty();
```
prathapreddy123, line 364: Can be simplified as 
`return input != null && !RecordWithMetadataHelper.of(input).getValue().isEmpty(); `

examples/dataflow-csvio/src/main/java/com/google/example/ExampleReadOptions.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Description("Sink File Pattern Blob")
+  @Required
+  public String getSink();
+
+  public void setSink(String value);
+
+  @Description("Path to quarantine of invalid records")
+  @Required
+  public String getQuarantine();
```
prathapreddy123, line 39: As InvalidRecordsSink or deadLetterSink are commonly used compared to Quarantine worth considering  changing to familiar name

examples/dataflow-csvio/src/main/java/com/google/example/ExampleReadOptions.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Description("Source File Pattern Blob")
+  @Required
+  public String getSource();
+
+  public void setSource(String value);
+
+  @Description("Sink File Pattern Blob")
+  @Required
+  public String getSink();
```
prathapreddy123, line 33: Is there any reason why getSink will have a file pattern instead of output directory path?

prathapreddy123 (commented):
@damondouglas -  Thanks for your contribution to address the challenges handling CSV variations.   Included few comments for your review.

examples/dataflow-csvio/src/main/java/com/google/example/csvio/CSVRowUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+/** Helpers to process CSV data into its expected {@link Schema.Field} types. */
+class CSVRowUtils {
+
+  static Row csvLineToRow(CSVFormat csvFormat, String header, String line, Schema schema) {
+    try {
+      CSVParser headerParser = CSVParser.parse(header, csvFormat);
+      CSVParser lineParser = CSVParser.parse(line, csvFormat);
+
+      List<CSVRecord> headerRecordsRaw = headerParser.getRecords();
```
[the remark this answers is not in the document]
damondouglas: The reason for downstream checks instead of bundling is that certain variables cannot be instantiated until previous variables are checked.  This is based on the fact that CSVRecord holds a list of "lines".

For example,

1) headerParser.getRecords() returns a List<CSVRecord> from which we need to check that there is only one header CSVRecord

2) Then, we need to evaluate the CSVRecord to check whether its field count matches the schema field count.

damondouglas:
@prathapreddy123 Thank you again for reviewing.  I've updated per most of the comments and this is ready for your review.

examples/dataflow-csvio/src/main/java/com/google/example/csvio/CSVRowUtils.java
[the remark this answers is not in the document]
prathapreddy123: Yes I understand that we need 2 checks for both header & line but my comment was about ordering of these checks as shown below:
```
 List<CSVRecord> headerRecordsRaw = headerParser.getRecords();

   if (headerRecordsRaw.size() != 1) {
        throw new IllegalArgumentException(
            String.format("header should be a single CSVRecord, got: %d", headerRecordsRaw.size()));
      }

      CSVRecord headerRecord = headerRecordsRaw.get(0);

 if (headerRecord.size() != schema.getFieldCount()) {
        throw new IllegalArgumentException(
            String.format(
                "mismatch of header size and schema size. header: %d vs schema: %d",
                headerRecord.size(), schema.getFieldCount()));
      }
      
      List<CSVRecord> lineRecordsRaw = lineParser.getRecords();
      if (lineRecordsRaw.size() != 1) {
        throw new IllegalArgumentException(
            String.format("line should be a single CSVRecord, got: %d", headerRecordsRaw.size()));
      }

      CSVRecord lineRecord = lineRecordsRaw.get(0);

      if (lineRecord.size() != schema.getFieldCount()) {
        throw new IllegalArgumentException(
            String.format(
                "mismatch of line size and schema size. line: %d vs schema: %d",
                lineRecord.size(), schema.getFieldCount()));
      }
```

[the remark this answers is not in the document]
damondouglas: Ah! I understand now thank you.

damondouglas:
This has been in the backlog for too long.  I will submit to the Beam project.  Thank you, again.

PR #866, opened 2022-07-28
Added java code that is used in a Google Cloud blog post. 

eenclona-google:
This PR add Cloud Support API code that is used in a Google Cloud blog post and added another example app to showcase its usage.


examples/cloud-support/java_update_cases_blog_code/my-app/src/main/java/com/mycompany/app/App.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.List;
+
+// starter code for cloud support API to initialize it
+public class App {
+
+  // Shared constants
+  static final String CLOUD_SUPPORT_SCOPE = "https://www.googleapis.com/auth/cloudsupport";
+
+  // TODO: update PROJECT with your project number
+  static final String PROJECT = "projects/123456789";
```
juliocc, line 39: can you make this some kind of input instead of hard coding it?

examples/cloud-support/java_working_app_example/README.md
```diff
@@ -0,0 +1,21 @@
...
+This use case solves the common problem faced by orgs with many support cases needing an update of the email addresses attached to their cases. For example, if an org creates a new email address to track all support cases and it wishes to add this email to all open support cases. This can be done easily in a programmatic way. Additionally, I also showed how to use the other methods exposed by the API in the code, like listing cases, getting a case, etc. 
+
+Please note that this sample app is not code complete - there may be some edge cases that were not accounted for. So please verify the completeness of this code before using this in production environment.
+
+## How to get started
+
+1. Make sure you have a support enabled GCP account.
+2. Create a service account that has the Organization Viewer and Tech Support Viewer role.
+3. Download the service account key
+4. Run the following command in the terminal: export GOOGLE_APPLICATION_CREDENTIALS=".../[path to service account key/.json"
```
juliocc, line 18: It seems you're using application default credentials. Do you really need a service account to run this code, specially in a section titled "How to get started"?

Perhaps you can simplify these instructions to use `gcloud auth login` and include a note that in a production environment you would use a service account with a link explaining different options to authenticate (keys, vm identity, WIF, etc)

juliocc (changes requested):

examples/cloud-support/java_update_cases_blog_code/my-app/src/main/java/com/mycompany/app/App.java
eenclona-google: It's definitely possible, but I think for the purposes of the blog post it's much easier to convey that they just need to update this variable instead of giving instructions on how to pass it in and all that. Also, the goal of this code is to showcase the capability of Cloud Support API, so I think having this hard coded doesn't make a big difference. I hope this is okay, but let me know!

examples/cloud-support/java_working_app_example/README.md
eenclona-google: That's a good point! Doing gcloud auth login is definitely a way to authenticate, but for the purposes of the blog post I'm trying to keep it consistent to the instructions posted in the user guide for Cloud Support API - https://cloud.google.com/support/docs/reference/v2#getting-started - which also talks about using service accounts. I hope this is okay as well! Thanks so much!

examples/cloud-support/java_update_cases_blog_code/my-app/src/main/java/com/mycompany/app/App.java
eenclona-google: Ah that makes sense! Thank you for the feedback. Will push an update to resolve this :)

examples/cloud-support/java_update_cases_blog_code/my-app/src/main/java/com/mycompany/app/App.java
```diff
@@ -0,0 +1,153 @@
...
+
+    try {
+      CloudSupport cloudSupport = getCloudSupportService();
+      // with the CloudSupport object, you may call other methods of the Support API
+      // for example, cloudSupport.cases().get("name of case").execute();
+
+      System.out.println("CloudSupport API is ready to use: " + cloudSupport);
+
+      projectResource =
+          projectResource + JOptionPane.showInputDialog("Please enter your project number: ");
```
juliocc, line 51: Opening a dialog forces the user to have a GUI to run your application. A better option is to pass a CLI argument

juliocc (approved):

PR #874, opened 2022-08-17
Add an example for workload identity federation

yuryninog:
Add an example for integrating Gitlab and Google Cloud using Workload Identity Federation.


agold-rh (approved):
LGTM

PR #877, opened 2022-08-23
BigQuery z/OS Mainframe Connector: add shim to enable compilation without IBM JZOS jars

jasonmar:
This PR adds a subpackage to the mainframe connector that allows users to compile against the shim package and avoid the need to download proprietary IBM jars from their z/OS to be able to build the mainframe connector.


tools/bigquery-zos-mainframe-connector/README.md
```diff
@@ -161,9 +162,9 @@ gsutil cp --replace --pic_t_charset="UTF-8" --remote=false gs://<my-bucket>/tabl
 After downloading the repo, build jar:
 ```
 cd professional-services/tools/bigquery-zos-mainframe-connector
-(cd  mainframe-util ; sbt "clean;publishLocal")
-(cd  gszutil ; sbt "clean;assemblyPackageDependency" )
-(cd  gszutil ; export appjar=true ; sbt "assembly" )
+(cd  jzos-shim; sbt publishLocal)
+(cd  mainframe-util; sbt publishLocal)
+(cd  gszutil; sbt assemblyPackageDependency; sbt package)
```
franklinWhaite, line 167: Just out of curiosity, why is it that in the `Test connector locally` section in line 141 we are just running `sbt assembly` while here we running `sbt assemblyPackageDependency; sbt package`?

franklinWhaite (approved):
LGTM!

tools/bigquery-zos-mainframe-connector/README.md
jasonmar: This can probably use a separate explanation section, because either way works and there's not a huge difference.
Generally, assemblyPackageDependency is used during development when the user needs to recompile the jar several times. I still wanted to retain one use of assemblyPackageDependency in the examples, I just didn't take time to break out a separate explanation of why it's used.

jasonmar: It's really helpful during development but for users that only need to build once, it's much easier to just use sbt assembly and be done with it.

PR #884, opened 2022-09-07
Mainframe Connector: remove inlining compiler options

jasonmar:
recently added jzos-shim method stubs must not be inlined by compiler, else IBM provided jzos libraries will not be used at runtime

This pull request removes compiler options that cause inlining of methods that must be provided by IBM libraries, resolving an issue which prevent the connector from working correctly when compiled against the recently added jzos-shim module.


franklinWhaite (approved):
LGTM

PR #888, opened 2022-09-13
Remove invalid pip instructions - security fix

agold-rh:
This removes obsolete / invalid pip instructions invoking an non-existent pypi package.
This was flagged as a security risk; the fix is to remove the invalid instructions.


aymanfarhat (approved):
LGTM

PR #889, opened 2022-09-14
Update setup instructions to reflect current CAI asset creation mode

apsureda:
CAI feeds are now created via terraform and not via shell scripts as it was explained in the README.


aymanfarhat (approved):
LGTM

PR #891, opened 2022-09-21
Mainframe Connector: add copybook support for unsigned int fields

jasonmar:
This PR adds support for parsing of additional copybook field types. It also enhances documentation of environment variables to increase initial success rate of installation.

Examples of new field types: 
PIC XXX.    3 byte character string
PIC 999.     3 byte character string containing unsigned integer
PIC 9(6)V9(2).    8 byte character string containing unsigned decimal with scale 2


franklinWhaite (approved):
LGTM!

PR #897, opened 2022-10-02
Add example of cloudbuild with docker-compose running tcp proxy in th…

yyzh1:
Add example of cloudbuild with docker-compose running tcp proxy in the background for all build steps


benking84 (approved):
LGTM

ryanmcdowell (approved):

PR #898, opened 2022-10-05
Added a flowchart and some text to explain how ranges are allocated.

cgrotz:
And removed the populate example data SQL migration step.


aymanfarhat (approved):
LGTM, thanks for the update!

PR #902, opened 2022-10-06
Mainframe Connector: bq extract

jasonmar:
This PR enables `bq extract` to accept control characters such as `DLE` and hex values such as `0x10` via the `--field_delimiter` command-line option.

### DIFF
--- a/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/ExtractOptionParser.scala
+++ b/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/ExtractOptionParser.scala
@@ -51,14 +51,35 @@ object ExtractOptionParser
     .action((x,c) => c.copy(format = x))
 
   opt[String]("field_delimiter")
-    .text("(optional) For CSV exports, specifies the character that marks the boundary between columns in the output file. The delimiter can be any ISO-8859-1 single-byte character. You can use \\t or tab to specify tab delimiters.")
+    .text("(optional) For CSV exports, specifies the character that marks the boundary between columns in the output " +
+      "file. The delimiter can be any ISO-8859-1 single-byte character. " +
+      "You can use '\\t' or 'tab' to specify tab delimiters. " +
+      "Control characters common to both ebcdic and ascii such as 'dle' are also accepted. " +
+      "You can also specify hex values such as '0x10'.")
     .action {(x, c) =>
-      if (x.equalsIgnoreCase("dle")) {
-        c.copy(delimiter = "\u0010")
-      } else {
-        c.copy(delimiter = x)
+      val delimiter = x.toLowerCase() match {
+        case "nul" => "\u0000"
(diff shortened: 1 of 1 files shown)


tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
```diff
@@ -0,0 +1,151 @@
...
+        val rowCount = BQ.rowCount(bq, table)
+        val result = extract(table, bq, cfg, zos)
+
+        // Publish results
+        if (cfg.statsTable.nonEmpty) {
+          val statsTable = BQ.resolveTableSpec(cfg.statsTable, cfg.projectId, cfg.datasetId)
+          val jobId = BQ.genJobId(cfg.projectId, cfg.location, zos, "extract")
+          val tblspec = s"${statsTable.getProject}:${statsTable.getDataset}.${statsTable.getTable}"
+          logger.debug(s"Writing stats to $tblspec, with jobId ${jobId}")
+          StatsUtil.retryableInsertJobStats(zos, jobId, bq, statsTable, jobType = "export", recordsOut = result.activityCount)
```
franklinWhaite, line 98: Seems to me this code is the same as in the previous if block

tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
```diff
@@ -0,0 +1,151 @@
...
+    } catch {
+      case e: Throwable =>
+        val msg = "Extract failed: " + e.getMessage + "\n"
+        logger.error(msg, e)
+        Result.Failure(msg)
+    }
+  }
+
+  def extract(table: TableId, bq: BigQuery, cfg: ExtractConfig, zos: MVS): Result = {
+    val jobId = BQ.genJobId(cfg.projectId, cfg.location, zos, "extract", Util.generateHashString)
```
franklinWhaite, line 115: why are we generating a custom BQ job id?

jasonmar: It's not strictly necessary because BigQuery will autogenerate an ID, but specifying our own makes it a human readable one that shows the z/OS job and step name.

tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
jasonmar: deduplicated the stats and logging but putting it into the extract method common to both query and table based extract

franklinWhaite:
"This branch is out-of-date with the base branch". Could you please update? @jasonmar

franklinWhaite (approved):

PR #903, opened 2022-10-07
Mainframe Connector: PDS upload

jasonmar:
This PR adds PDS upload to the `scp` command.

Example usage:
`scp PATH.TO.PDS gs://bucket/prefix/`
would copy each member of PDS to `gs://bucket/prefix/PATH.TO.PDS/MEMBERNAME`

`scp PATH.TO.PDS gs://bucket/prefix`
would drop the DSN from the object name and copy each member of PDS to `gs://bucket/prefix/MEMBERNAME`

If the input DSN references a PDS, EBCDIC to ASCII conversion will be enabled by default but this can be disabled by specifying `--noConvert`. Encoding can be specified by the `--encoding` command-line option.

### DIFF
--- a/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
+++ b/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2022 Google LLC All Rights Reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/cmd/Extract.scala
```diff
@@ -0,0 +1,151 @@
...
+      } else {
+        // Extract contents of a table
+        val table: TableId = BQ.resolveTableSpec(cfg.sourceTable, cfg.projectId, cfg.datasetId)
+        logger.info(s"Extract from source table: $table")
+        val rowCount = BQ.rowCount(bq, table)
+        val result = extract(table, bq, cfg, zos)
+
+        // Publish results
+        if (cfg.statsTable.nonEmpty) {
+          val statsTable = BQ.resolveTableSpec(cfg.statsTable, cfg.projectId, cfg.datasetId)
```
franklinWhaite, line 94: same comment as in https://github.com/GoogleCloudPlatform/professional-services/pull/902#discussion_r990410420

franklinWhaite:
@jasonmar 
LGTM, the only detail is that branch is out-of-date please update.

franklinWhaite (approved):

PR #915, opened 2022-10-19
Patch Go version in go.mod

damondouglas:
This PR patches [examples/dataflow-custom-templates/go/go.mod#L3](https://github.com/GoogleCloudPlatform/professional-services/blob/main/examples/dataflow-custom-templates/go/go.mod#L3) from `1.20` to `1.19` as the one specified is not yet released.


aymanfarhat (approved):
LGTM, thank you for your contribution.

PR #933, opened 2022-11-15
add examples/dataproc-job-optimization-guide

CYarros10:
This guide is designed to optimize performance and cost of applications running on Dataproc clusters.  Because Dataproc supports many big data technologies - each with their own intricacies - this guide is intended to be trial-and-error experimentation.  Initially it will begin with a generic dataproc cluster with defaults set. As you proceed through the guide, you’ll increasingly customize Dataproc cluster configurations to fit your specific workload.


agold-rh:
@CYarros10 Are you still looking to have this merged?

agold-rh (approved):

PR #935, opened 2022-11-22
Mainframe Connector date envvars, sql quoting, safe job id generation

jasonmar:
This PR includes the following changes to the mainframe connector:
- prevent BigQuery API failure due to disallowed character in job name or step name by filtering during job id creation
- fix issue preventing use of single quotes in sql provided by command-line
- add date and time environment variables

### DIFF
--- a/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/BQ.scala
+++ b/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/BQ.scala
@@ -21,9 +21,12 @@ import com.google.api.services.bigquery.{Bigquery, model}
 import com.google.auth.Credentials
 import com.google.auth.http.HttpCredentialsAdapter
 import com.google.cloud.bigquery.InsertAllRequest.RowToInsert
-import com.google.cloud.bigquery.{BigQuery, BigQueryError, BigQueryException, CopyJobConfiguration, DatasetId, ExtractJobConfiguration, Field, FieldList, InsertAllRequest, InsertAllResponse, Job, JobConfiguration, JobId, JobInfo, JobStatus, LoadJobConfiguration, QueryJobConfiguration, Schema, StandardSQLTypeName, StandardTableDefinition, TableDefinition, TableId}
-import com.google.cloud.imf.gzos.pb.GRecvProto
-import com.google.cloud.imf.gzos.{MVS, Util}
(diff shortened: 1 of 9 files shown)


tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/BQ.scala
```diff
@@ -38,10 +38,15 @@ object BQ extends Logging with GoogleApiL2Retrier {
   override val retriesTimeoutMillis: Int = Services.l2RetriesTimeoutMillis
 
   def genJobId(projectId: String, location: String, zos: MVS, jobType: String, suffix: String): JobId = {
```
franklinWhaite, line 40: There is a [method in this class](https://github.com/jasonmar/professional-services/blob/mainframe-connector-a/tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/BQ.scala#L57) with the same name and slightly different signature. Should't we also update that one to disallowed character in job name or step name by filtering during job id creation?

tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/Bqsh.scala
```diff
@@ -184,8 +217,10 @@ object Bqsh extends Logging {
         if (!(singleQuoted || doubleQuoted)) {
           if (c == '\'') {
             singleQuoted = true
+            sb.append(c)
```
franklinWhaite, line 220: The test [it should "replace variables" in](https://github.com/jasonmar/professional-services/blob/mainframe-connector-a/tools/bigquery-zos-mainframe-connector/gszutil/src/test/scala/com/google/cloud/bqsh/ShellSpec.scala#L99) Seems to have been broken by this change. 

Actually, If my interpretation is correct, I would say the test should be modified to reflect the current behavior of this function since `replaceEnvVars` no longer removes quotes.

jasonmar: updated the test

tools/bigquery-zos-mainframe-connector/gszutil/src/main/scala/com/google/cloud/bqsh/BQ.scala
jasonmar: consolidated into a single method with optional random suffix flag

franklinWhaite (approved):
LGTM!

PR #953, opened 2022-12-13
Mainframe Connector: add curl command

jasonmar:
This PR adds a curl command which is meant to be used to call REST APIs such as Airflow DAG launch or Elasticsearch index add.


franklinWhaite:
LGTM!
Just thinking that as we add more command we might want to add some documentation on them, i.e. example JCL, help text. Anyhow this could be a separate PR

PR #968, opened 2023-01-04
add support for signed numeric character fields

jasonmar:
This PR adds support for encoding and decoding signed numeric character fields.

Examples:
PIC +9(3). should encode 123 as '+123'
PIC +9(7).9(4). should encode 123 as '+0000123.0000'


franklinWhaite:
LGTM

PR #981, opened 2023-01-27
Add Capacity Planner Sheets Extension

busunkim96:
This is the Google Sheets extension version of https://github.com/GoogleCloudPlatform/professional-services/pull/975. 

#975 should be merged first as this PR has some text that depends on that existing first.

*Style Guide*: I tried to align to the https://google.github.io/styleguide/jsguide.html, but have likely missed things as JavaScript isn't my usual language. I thank you in advance for your patience, reviewers. 🙏 

*Unit Tests*:I tried and failed to figure out how to write unit tests (or even `assert`s) that I could run in the Apps Script editor. If someone is able to point me to something I will write those tests. 😄 

Other notes:

* Files are saved as `.js` even though `.gs` is the extension that is used for Apps Script. This was mainly for correct syntax highlighting in my code editor (and probably in GitHub?)
* Instructing the user to upload the XLSX then convert is weird and inconvenient. I did it because it's hard for me to directly share a Google Sheet publicly from my work account. If you have better ideas, please let me know. 

FYI @shunjikawabata


tools/capacity-planner-sheets-extension/queries.js
```diff
@@ -0,0 +1,136 @@
...
+ * 
+ * This file stores the queries for the Google Cloud Monitoring API.
+ * 
+ * NOTE: Apps Script only allows HTML or .gs files so there is no simple way to include a TOML or JSON file here.
+  Queries are originally from the queries.toml file below.
+
+  [TODO: link to Capacity Planner CLI TOML file]
+*/
+
+const QUERIES = {
```
aymanfarhat, line 24: Looks good. Just an idea would be good to consider generating this file based on the toml configuration in the other repo. Would be good as well to add a link to the other repo when it is up I agree.

tools/capacity-planner-sheets-extension/sheets.js
```diff
@@ -0,0 +1,185 @@
...
+    // Gap calculations
+    sheet.getRange(`M4:M${lastRowIndex}`).setFormula(`IF( AND(J4 > 0, K4 > 0), K4 - J4, "NA")`);
+    sheet.getRange(`N4:N${lastRowIndex}`).setFormula(`IF( AND(J4 > 0, K4 > 0), M4/J4, "NA")`);
+
+}
+
+/**
+ * Formats the sheet to look like the template.
+ * This is hard to understand as code, reference the template sheet for the intended output.
+ * https://docs.google.com/spreadsheets/d/1OpcMZnuPgV2InHlxybv6DaOmzvUMRx5eC-KzL1UM_Ok/edit?resourcekey=0-Wr1Pi9GbOmnwaJVckGXE0w#gid=623747450
```
aymanfarhat, line 110: This sheet doesn't seem to be public (I don't have access). Do you need to reference it given that you have the initial template as an excel file? Or is this sheet file you're referencing different ?

tools/capacity-planner-sheets-extension/sheets.js
```diff
@@ -0,0 +1,185 @@
...
+function formatSheet_(sheet) {
+    let lastRowIndex = sheet.getLastRow();
+
+    // Merge some header cells
+    ["A1:D3", "E2:E3", "K2:O3"].forEach(range => sheet.getRange(range).mergeVertically());
+    ["E1:I1", "J1:O1"].forEach(range => sheet.getRange(range).mergeAcross());
+
+    // Font colors and styles
+    sheet.getRange("J2:O3").setFontColor("#cfe2f3");
+    sheet.getRange("E4:I").setFontColor("#666666");
```
aymanfarhat, line 123: Given that these hex color codes are repeated  used across the script, might be more maintainable to store them into a simple object e.g. 

```
const colors = {
      "lightBlue": "#cfe2f3",
      "darkGrey":  "#666666"
};

sheet.getRange("J2:O3").setFontColor(colors.lightBlue);
```

Naming convention on keys would depend on your use case of course, not necessarily color names but whatever you see fit.

tools/capacity-planner-sheets-extension/queries.js
```diff
@@ -0,0 +1,136 @@
...
+  "bigtable": {
+    "product_name": "Cloud BigTable",
+    "metrics": {
+      "egress": {
+        "metric_name": "Egress MB/s",
+        "query": "fetch bigtable_table\n        | metric 'bigtable.googleapis.com/server/sent_bytes_count'\n        | align rate(1m)\n        | every 1m\n        | group_by [], [value_sent_bytes_count_aggregate: aggregate(value.sent_bytes_count)]\n        | scale ('MiBy/s')"
```
aymanfarhat, line 30: I noticed there is a lot of extra spacing in this file such as this line and other lines. Is this spacing needed or was it create by mistake?

tools/capacity-planner-sheets-extension/sheets.js
```diff
@@ -0,0 +1,185 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * 
+ * 
+ * Convenience functions for creating, writing to, and formatting sheets.
+ */
+
+BACKGROUND_COLORS = ["#ead1dc", "#cfe2f3", "#d0e0e3", "#fff2cc", "#f4cccc"];
```
aymanfarhat, line 20: This variable is not used, consider removing it if that's the case. If you're keeping it, consider setting it as `const`

tools/capacity-planner-sheets-extension/sheets.js
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Creates a new sheet with the given name.
+ * NOTE: If a sheet with sheetName already exists, this function will delete it and create a new
+ * sheet with the same name.
+ * @param {string} The name of the new sheet
+ */
+function createSheet_(sheetName){
+  let ss = SpreadsheetApp.getActive();
```
aymanfarhat, line 29: For variables who's values won't be re-assigned consider using `const` instead of `let`. When using const, the value it holds *can still be mutable* e.g. `ss.deleteSheet(sheet);` but the identifier cannot be re-assigned. I would suggest using it accordingly where you see that fit.

tools/capacity-planner-sheets-extension/README.md
```diff
@@ -0,0 +1,51 @@
+# Capacity Planner Sheets Extension
+
+This tool is a Google Sheets extension for capacity planning on Google Cloud. 
+This tool extracts peak resource usage values and corresponding timestamps for a given Google Cloud project, time range and timezone.
+
+
+## Setup
```
aymanfarhat, line 7: As long as you get the desired results, the .xlsx upload as a template approach should work okay. I understand that making a google sheet as a template public might not be a good idea. Another possibility (potentially future) would be have an init function step (implemented in the same app script) that initialises a given new sheet with the placeholders needed before usage. But this would involve adding more code as well, so it's really up to you and whatever works for your users.

aymanfarhat (changes requested):
Looks good overall, added some points to look into before merging. Thanks!

tools/capacity-planner-sheets-extension/sheets.js
busunkim96: The excel should be referenced instead, I will update it.

tools/capacity-planner-sheets-extension/queries.js
busunkim96: @shunjikawabata Do you know if the monitoring API cares about the extra spaces?

busunkim96:
Thanks for the review @aymanfarhat! I forgot I'd disabled GitHub email notifications and completely forgot about this PR. 🤦‍♀️  I'll follow up on your comments next week.

tools/capacity-planner-sheets-extension/queries.js
busunkim96: Updated to link to the CLI and added a Python script for converting from toml.

tools/capacity-planner-sheets-extension/README.md
busunkim96: Thanks for the advice @aymanfarhat!  I unfortunately don't have additional cycles to spend on this right now, so I'll leave it like this for now. 

If I end up getting the time/budget to do more work on this I'd definitely like to add an auto-init function.

busunkim96 (commented):
Sorry for disappearing for so long, this is ready for re-review @aymanfarhat 🙏

tools/capacity-planner-sheets-extension/sheets.js
busunkim96: Thanks for the explanation! I've removed all instances of `var` and switched to `let` and `const.

tools/capacity-planner-sheets-extension/queries.js
busunkim96: It looks like the spaces are from the TOML (where the spaces are actually line breaks). The conversion script strips away the extra spaces now.

tools/capacity-planner-sheets-extension/sheets.js
busunkim96: Done

tools/capacity-planner-sheets-extension/sheets.js
busunkim96: Updated comments like this to either point to the Capacity Planner CLI (in this repo) or tell the user to look at a sample sheet in the XLSX.

tools/capacity-planner-sheets-extension/README.md
aymanfarhat: ACK 👍

aymanfarhat (approved):
LGTM

aymanfarhat:
@busunkim96 Changes approved and merged, thank you for your contributions!

PR #1001, opened 2023-03-09
bigquery-translation-validator-utility added

aaslam17:
This utility helps in the validation of the SQL by comparing both the Legacy Source SQL and Bigquery SQL ,once the translation is completed.
This utility helps to pin-point any major translation errors by performing validations like table, column, join and function validation and provide a report with the validation results.
This utility will be of major help to quickly identify translation errors, right at the beginning of the migration rather than figure out the issue later through data validation and hence saving a lot of time for the migration projects.


shadowshot-x:
cc: @rishabkhawad @stunner-07

PR #1009, opened 2023-04-06
bigquery-long-running-optimization added

shadowshot-x:
A utility reads the entire SQL and provides a list of suggestions that would help to optimize the query and avoid the long running issues.
The utility will assist to prevent 2 major cases
1. The  Queries go into long running mode and eventually fail with the timeout error after running for 6+ hours. 
2. The query runs for many hours , consumes more computation and finally produces an output.


shadowshot-x:
cc: @aaslam17 @haprcr @ tjatGoogle

agold-rh (approved):

PR #1012, opened 2023-04-14
GKE custom organizational policies based on CIS Benchmarks

athankso60:
This repository is an attempt to provide sample organizational policies that could be enforced in order to comply with the [GKE portion](https://cloud.google.com/kubernetes-engine/docs/concepts/cis-benchmarks#default_values_on) of CIS benchmarks.


tools/gke-custom-org-policies/run.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/sh
```
shakeebshams, line 1: the script contains arrays and `[[` which are available in bash. The shebang line should be `#!/bin/bash`

tools/gke-custom-org-policies/run.sh
```diff
@@ -0,0 +1,123 @@
...
+        cd $value
+            content=($(ls -d */))
+            for policy in ${content[@]} 
+            do
+            cd $policy
+                files=($(ls))
+                echo ${files[0]}
+                #run constraint
+                constraint=`cat ${files[0]}`
+                echo "${constraint/"[ORGANIZATIONID]"/$1}"  > constraint.yaml
```
shakeebshams, line 37: this is another bash line that may not work in a bourne shell

shakeebshams (commented):
some suggestions

agold-rh:
Closed as stale. Please re-open if I'm wrong.

PR #1018, opened 2023-04-24
Added support for mixed case and special characters in schema and tab…

simonpane:
Existing scripts will fail if the schema and/or table name use special characters or are in lower case.  Solution is to use Oracle quoted identifiers.

Example source schema to analyze:
```SQL
create user "simon test" identified externally quota 1m on users;
grant connect,resource to "simon test";

create table "simon test"."table one" (c1 number);
insert into "simon test"."table one" values (1);
commit;
```

Corresponding variable adjustments to make (example for file `Oracle_Number_Recommender_GCP.sql` lines 72-75):
```SQL
    --Add schema name to be scan for number data type recommender.
    schemaname varchar2(4000) := 'simon test'; --'SCHEMA';
    --Add semicolon(;) separated list of table name or add wildcard % to scan all table in the specified schema.
    tablename varchar2(4000) := '%';  --'%'; 
```

Without changes from this PR, script will fail to select and process the `"simon test"."table one"` table.  With the changes from this PR, the table will be properly selected and processed.

NOTE: documentation should possibly be updated accordingly, as the **schemaname** and **tablename** varaibles, if explicitly defined, will need to use the proper case since the `UPPER()` function must be removed.


agold-rh (commented):
@simonpane Could you add a newline to the end of each file? It will silence a visual warning.

simonpane:
Reverted newlines at end of file @agold-rh .  Although please note that the visual warning was due to the original commit (before my changes) not having a newline. My commits added the newline at EOF.  Regardless, I have reverted that part of my change via PR c5195d3923c4f5f5c01e26bf0906c2581e72cbbc in this PR.  Thank you.

agold-rh (approved):

PR #1022, opened 2023-05-02
Google Cloud Support Slackbot - README and Dockerfile updates

TheLanceLord:
Updated README.md to fix deprecated calls used to build the Cloud Run app. Updated the Dockerfile for a slightly more compressed image size.


TheLanceLord:
Discovered issues where app engine is automatically creating Firestore in Datastore mode rather than Firestore Native mode. Need to adjust code to fix this

PR #1032, opened 2023-05-29
Cloud Composer Test Dag Generator

haris-aryan:
## Cloud Composer test DAG generator 

This PR is raised to add test DAG generator. This utility provides a way to generate large number of DAGs with variable number of tasks using different operators. It will be helpful for Cloud Composer load simulation and also fine tuning environment configs based on composer metrics.


alexshires:
@haris-aryan LGTM - who is an approver here?

iht (approved):
Approving on behalf of Alex

PR #1038, opened 2023-06-13
Remove Fabric fork

juliocc:
Remove Vertex AI MLOps example that copied most of the [Fabric/FAST repository](https://github.com/GoogleCloudPlatform/cloud-foundation-fabric).


ludoo (approved):
Approving, since we've been trying to engage with the original author for a couple of months, offering support to make this a proper fit to the current version of FAST instead of a hacked clone of an old version (with wired-in defaults too), and saw zero traction.

PR #1058, opened 2023-08-15
GCP Slackbot - Added automatic tracking of cases; fixed issue with escalations breaking case updates

### LINKED ISSUE #1037: Google Cloud Support Slackbot breaks with case escalations
/GoogleCloudPlatform/professional-services/google-cloud-support-slackbot/support_case.py for @TheLanceLord 

When fetched with the Cloud Support API, escalated cases aren't returning the displayName field in the Case Object. Having a Support Engineer or member of the account team go into the escalated case and adding a name to the field and then redeploying the bot fixes the issue. A longterm fix is needed to:

1. Check that fields exist in the returned Case object and assign them empty values if they don't
2. Nest the case updates process in a while loop with a try condition so it can auto-heal if there is an issue without the need to redeploy the bot

TheLanceLord:
Added functionality for automatically tracking cases that meet criteria set by a user, specifically an asset to track like a specific organization, folder, or project, and the priorities to automatically track against, e.g. P1, P2. The next release will update the functionality so that cases that go through a priority change, or cases that are escalated, will also become tracked. Updated the integration test to include the new functions and provide better details on the results of each test.

Closes #1037 by surrounding the attempt to get the display name for the case with a try-except block.

Code was linted with flake8


tools/google-cloud-support-slackbot/autotrack_edit.py
```diff
@@ -0,0 +1,111 @@
...
+                  f" tracker: \n /google-cloud-support autotrack-create"
+                  f" {asset_type} {asset_id} P1 ... P4")
+        )
+
+
+if __name__ == "__main__":
+    test_channel_id = os.environ.get("TEST_CHANNEL_ID")
+    test_channel_name = os.environ.get("TEST_CHANNEL_NAME")
+    test_user_id = os.environ.get("TEST_USER_ID")
+    test_project_id = os.environ.get["TEST_PROJECT_ID"]
```
shakeebshams, line 108: this code should be 
```suggestion
    test_project_id = os.environ.get("TEST_PROJECT_ID")
```

tools/google-cloud-support-slackbot/list_autotrack_all.py
```diff
@@ -0,0 +1,79 @@
...
+        PROJECT_ID = os.environ.get("PROJECT_ID")
+        cred = credentials.ApplicationDefault()
+        firebase_admin.initialize_app(cred, {
+            "projectId": PROJECT_ID,
+        })
+
+    db = firestore.client()
+    collection_name = "auto_case_tracker"
+    doc = f"{collection_name}/{channel_id}"
+    channel_tracking = db.document(doc).get()
```
shakeebshams, line 54: here you are retrieving the doc reference and not the actual data. to get the data do 

```suggestion
    channel_tracking = db.document(doc).get().to_dict()
```

TheLanceLord: we get the data later on in line 55 with ` asset_types = channel_tracking.reference.collections()` and then line 59 `       tracked_assets = asset_type.get()`

shakeebshams (approved):

agold-rh (approved):

PR #1089, opened 2023-09-06
add cloud-composer-migration-complexity-assessment

CYarros10:
An Airflow DAG that uses a variety of tools to analyze a Cloud Composer 1 environment, determine a work estimate, and accelerate the conversion of airflow 1 dags to airflow 2 dags.


tools/cloud-composer-migration-complexity-assessment/airflow-v1-to-v2-migration/misc/migration.py
```diff
@@ -0,0 +1,169 @@
...
+        for root,dirs,files in os.walk(self.input_dir):
+            for file in files:
+                if file.endswith(".py"):
+                    filepath =  root + "/"+file
+                    new_dir = self.output_dir+root[inp_len:]
+                    isExist = os.path.exists(new_dir)
+                    if not isExist:
+                        os.makedirs(new_dir)
+                    new_file = new_dir+"/"+file
+                    change_count = 0
```
shakeebshams, line 75: why is `change_count` being re-setted here? This is then counting changes per file and not how many changes in total (if that is not the intended behaviour)

CYarros10: Yes this intended behavior - change count determines if the file will be considered "impacted"

dhaval-d (approved):

PR #1093, opened 2023-09-14
add tools/airflow-states-collector - Collect and push airflow task states to BQ

nikunjbhartia:
This tool creates an Airflow Dag which incrementally collects DagRun and TaskInstance metadata information and stores the information to a BigQuery Table. 
The tool also creates a BQ View which de-duplicates data, and gets latest state for every taskinstance. Then it creates a LookerStudio Dashboard on the BQ view.

Both Airflow1 and Airflow2 are supported.

State Information Captured by the tool:
- DagRuns: DagId, runId, run_start_ts, run_end_ts
- TaskInstance: Id, job_id, operator, start_ts, end_ts, updated_ts


nikunjbhartia:
Will raise the PR from my corp account

PR #1127, opened 2023-10-12
IPAM Autopilot: Adapt subnet overlap check

wolfmayr:
Allow creation of new IP ranges, even if there is an overlap in the already existing ranges. 

Currently, if two identical or overlapping ranges are already in the IPAM database for a given source range, the `verifyNoOverlap` function always returns an error. This makes it impossible to add a new IP range, even if the requested new IP Range would still fit into the source range without overlap. 

For an example see my included test, where the IP range `10.0.4.0/22` is added twice. This can for example happen if someone (or in our case our crossplane provider) adds a parent range twice by mistake. Adding a parent range with a set CIDR uses the `directInsert` function, which means there is no check for overlap, and duplicate inserts can happen.

This leads to a situation, where every possible IP range which is checked by the `findNextSubnet` function is always denied, because the `verifyNoOverlap` function finds the overlap in the **existing** ranges, and returns an error, even if the actually checked new range would fit without overlap. 

I rewrote the `verifyNoOverlap` function in a way that allows for a new range to be added, even if there is an overlap between existing subnets.


wolfmayr:
@cgrotz Hey, would be great if you find the time to have a look on this PR. :)

cgrotz:
Thank you @wolfmayr, that's a very elegant and useful improvement. Thank you for your contribution.

cgrotz (approved):

PR #1142, opened 2023-10-23
Update Migration Assessment tooling to not include converted code in output

CYarros10:
We would like the migration assessment tool to be utilized for scoping and informational purposes, therefore we will prevent uploading converted code to the output bucket.


alexshires (approved):
lgtm

PR #1143, opened 2023-10-23
Update Cloud Composer Example Tests with new DAG Quality test

CYarros10:
Tests all DAGs within an input DAG path directory.

Follow [Google Cloud Blog: Optimize Cloud Composer via Better Airflow DAGs](https://cloud.google.com/blog/products/data-analytics/optimize-cloud-composer-via-better-airflow-dags) - a guide containing a generalized checklist of activities when authoring Apache Airflow DAGs. 

test_dags_quality.py - enforces the following best practices from the guide above and provides many deprecation warnings:

* No Import Errors
* Valid Schedule Interval
* Owner Present
* SLA Present
* SLA Less Than Timeout
* Retries Present and value = 1-4
* Catchup Set to False by Default
* DAG Timeout set
* DAG Description set
* DAG Paused on Create
* DAG Valid Tags
* DAG Check Task Cycle
* DAG Import Time Less Than 2 Seconds

### DIFF
--- a/examples/cloud-composer-examples/tests/test_dags_quality.py
+++ b/examples/cloud-composer-examples/tests/test_dags_quality.py
@@ -1,0 +1,151 @@
+# Copyright 2018 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the u"License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


alexshires (approved):
thanks, lgtm

PR #1162, opened 2023-10-26
Fix requirements.txt after dependabot broke compatibility

javiergp:
Fix requirements.txt after dependabot broke compatibility


svidiella (approved):
LGTM

PR #1282, opened 2024-05-11
Add a new example: Redis Cluster Stateful MIG

shunjikawabata:
This example is to deploy high availability Redis clusters on Compute engine. It assumes that you're familiar with the reference architecture in High availability stateful clusters on Compute Engine.


examples/redis-cluster-stateful-mig/consul/consul-cluster/consul-cluster/main.tf
```diff
@@ -0,0 +1,257 @@
...
+    google_compute_instance_template.consul_server
+  ]
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+# Create the Instance Template that will be used to populate the Managed Instance Group.
+resource "google_compute_instance_template" "consul_server" {
```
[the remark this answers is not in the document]
bmenasha: dnsmasq seems to respond only when listen-address contains the forwarding rule ip?   

listen-address=10.128.0.32

i couldn't get it to work otherwise.

examples/redis-cluster-stateful-mig/redis/startup_script.sh
```diff
@@ -0,0 +1,265 @@
...
+  done
+}
+
+configure_and_join_consul_cluster() {
+  my_ip=$1
+  echo "Configuring the consul agent"
+  apt-get install consul -y
+  cat > /etc/consul.d/redis-master-role.sh <<EOF
+#!/bin/bash
+if [[ master == \$(redis-cli -h $my_ip info replication | tr -d '\r' | grep role: | cut -d : -f2 ) ]]; then echo 'matched'; else slave; fi
```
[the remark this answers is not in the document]
bmenasha: oh, i see it should be "else exit -1"   as we do want to error with a non-zero exit status.

examples/redis-cluster-stateful-mig/README.md
```diff
@@ -0,0 +1,204 @@
+# Deploy high availability Redis clusters on Compute Engine
+
+This example is to deploy high availability Redis clusters on Compute engine. It assumes that you're familiar with the reference architecture in High availability stateful clusters on Compute Engine.
+
+## Architecture
+There are several patterns for configuring Redis for high availability (HA). Redis Sentinel and Redis Cluster are the most common. This design focuses only on Redis Cluster. For Redis Sentinel use cases it is recommended to use Memorystore for Redis. If you need zone awareness in cluster management for better operability or other capabilities like auto-tiering, consider Memorystore for Redis Cluster or Redis Enterprise by Redis Inc.
+
+![Redis Architecture](./assets/redis_architecture.png "Redis Architecture")
+
+- At minimum, 1 set of 3 nodes consisting of a master node and its replicas is deployed to ensure high availability. If multiple sets representing shards are deployed, each master in sets should reside in distinct zones. This provides resilience against single VM failures and zonal failures.
```
bmenasha, line 10: What's the purpose of having one DNS name  for all primary and another for all replica nodes?

We are starting 6 nodes. And we configure the cluster with 1 replica for each primary.  This means there are 3 shards. 

Yet we only have two DNS names.   How does this work? 

Consul isn't shard aware, and even if it was, how would the client know which shard a key is in? 

That is, i want to get or set a key "X",   what server should i talk to?   Consul will return 3 hosts as the primary, but only one of them hosts the shard for key X, how do I know which one to use? 

It seems to me that in order to properly communicate to  Redis Cluster we MUST use the Redis Cluster protocol.  And if we are doing that, what's the point of having different DNS names for the primary and replica nodes? 

Thanks

agold-rh:
@shunjikawabata If you can address the comments by @bmenasha , I can get this merged.

shunjikawabata:
Thanks @agold-rh and @bmenasha for the comments/suggestions. We'll take a look and try fixing them. 
cc: @yashwantmahawar

examples/redis-cluster-stateful-mig/README.md
shunjikawabata: Yes, relying on the Redis Cluster protocol or a smart client/proxy is usually the most effective way to handle service discovery and shard routing in a Redis Cluster environment. The primary reason to having separate DNS names for primaries and replicas in this example is to showcase how the reference architecture in High availability stateful clusters on Compute Engine works with real stateful applications like a Redis Cluster.
Although there are tradeoffs, this architecture would be applicable to a massive scale Redis Cluster which requires manual sharding, meaning having multiple sets of Redis Clusters with a primary and replicas per a subset of entire key ranges for better performance. While less common, this approach can also offer some basic load balancing for read operations, especially in scenarios where you might want to direct some read traffic to replicas to offload primaries.
Probably these things should be clarified in the README. I'll update it.

bmenasha: >  relying on the Redis Cluster protocol or a smart client/proxy is usually the most effective way to handle service discovery and shard routing in a Redis Cluster environment. 

It's not a matter of efficiency, i just don't see how it works at all.


> this architecture would be applicable to a massive scale Redis Cluster which requires manual sharding, meaning having multiple sets of Redis Clusters with a primary and replicas per a subset of entire key ranges for better performance. 

Once you determine cluster to talk to, how do you choose the node within the cluster using  this example?   I can understand how this Consul/DNS based routing between primary and replica  will  work when the cluster has a single shard.   If so, that is a big limitation,  and something we should  document and explain and then of course change the example to only provision a single shard.  







Although there are tradeoffs, this architecture would be applicable to a massive scale Redis Cluster which requires manual sharding, meaning having multiple sets of Redis Clusters with a primary and replicas per a subset of entire key ranges for better performance.

I'm not saying the proposal is not efficient, I'm saying I don't understant how it works.

shunjikawabata: It will rely on Redis Cluster protocol once the client resolved the IP through DNS/Consul, and further communication will happen between the client and the node directly. Probably adding an example client with Redis configurations would help the readers to understand how it works?

bmenasha: thanks. i do think it will be helpful to show how clients connect. maybe just using redis-cli.
if Redis Cluster protocol is used,  what value is there in having two DNS names? 
thanks

agold-rh:
@shunjikawabata If you can address the comments by @bmenasha, I can get this merged.

PR #1299, opened 2024-06-12
Tool that converts slo definition to Terraform code

arjunvijaygoogle:
The tool basically converts the SLO Definition in GCP to its equivalent terraform block. Following are the exhaustive list of different SLO definition covered by this tool that can be converted

Request based SLI
Good Total Ratio
Distribution cut
Window Based SLI
Metrics Mean
Metrics Sum
Good Total Ratio Threshold
The tool has a very basic UI where user can provide the input of SLO definition and after clicking on convert button it will be converted to terraform block


arjunvijaygoogle:
@agold-rh Can you please review this

PR #1335, opened 2024-08-12
Application framework for redis performance benchmarks

prathapreddy123:
Application Framework to execute various operations on redis to evaluate key performance [metrics](https://cloud.google.com/memorystore/docs/redis/supported-monitoring-metrics) such as CPU & Memory utilization, Bytes transferred, Time per command etc. 

Framework currently supports few operations (set/get, push & pop) with sample payload. Payload is encoded in
[message pack](https://msgpack.org/index.html) format and compressed using [LZ4](https://lz4.org/) compression.

 Application can be extended to support additional redis operations, different encodings (e.g: avro, protobuf) and
 compression formats.


agold-rh:
@prathapreddy123 Are you planning on having @iht review this?

agold-rh (approved):

Issue #1374, opened 2024-10-23
[Asset Inventory] Dataflow pipeline is using an deprecated SDK version (2.41.0)


MarcFundenberger:
The GCP Dataflow console is alerting us that the Dataflow SDK version used by the Asset Inventory tool is deprecated (2.41.0).
Can you please modify the template to use an up-to-date SDK ?

bmenasha:
this pull request should resolve it. https://github.com/GoogleCloudPlatform/professional-services/pull/1394    
I'll update the latest pipeline after it's merged. 
thanks

bmenasha:
change is merged. dataflow template was updated. let me know if there are any problems.
thanks

MarcFundenberger:
Hello
Thanks for your work. Pipeline is working as intended.
However the column resource.json_data changed type from STRING to JSON with the new pipeline.
I agree it's better this way but it has the unforeseen consequence that you can't query it on dates that where generated with the old and new version of the pipeline...
E.g. : 
```
select resource.json_data from `dfdp-sre-data.shardedAdeoAssetInventory.container_googleapis_com_Cluster_202411*`
```
Crashed with error `Cannot read field of type STRING as JSON Field: resource.json_data`
But the column resource.json_data holds the same data than the rest of the column, so it's basically redundant... 
Anyway,  think it's bad practice to use our inventory Dataset this way, so I'll educate our users that try this (and complain) .

Perhaps next time we could work out a way for us tho test a pre-release version of the pipeline ? 

Thanks again.

bmenasha:
Yes, this was a backwards incompatible change.  When this code was originally written, BigQuery didn't have a JSON datatype.   Honestly if the JSON datatype existed back then I doubt we would have written this tool :) 

> Perhaps next time we could work out a way for us tho test a pre-release version of the pipeline ?

The Dataflow template  gs://professional-services-tools-asset-inventory/test/import_pipeline  is typically updated prior to the latest Dataflow template if you want to get changes sooner. 

> so it's basically redundant...

There are situations where the BigQuery schema can't represent the data,  for example when the bigquery column count is exceeded, the depth of the nested records exceeds bigquery's limits, the column name is too long.  We just truncate the data in such cases, but I didn't want to lose the original data in that event.

MarcFundenberger:
> We just truncate the data in such cases, but I didn't want to lose the original data in that event.

Yes, you're right, of course.
Thank you for your answers.

Issue #1373, opened 2024-10-23
[Asset Inventory] Dataflow stalled (over 24h instead of 3h) on full CAI Ressource Dump


MarcFundenberger:
We are using the tool "Asset Inventory" DataFlow Pipeline to load all of our Resources and IAM Rules in BigQuey Datasets.
Suddenly on the 18th of October 2024, the "resource" run of the DataFlow pipeline did not finish (we stopped it after more than 24h).  Prior to that date, typical running time was 3h.

The CAI dump is a full resource dump, meaning we do not specify any assetType to dump.

We tried removing the few new assetType  listed in CAI Release Notes from the list of assets found in CAI documentation.
That resulted in normal succesfull load, BUT the CAI Dump file is way smaller (60GiB instead of 70GiB for a dump), which indicates that we are missing some important assetTypes...

Is there any way to make sure that the Dataflow pipeline can handle a "full" CAI resource dump ?

bmenasha:
Marc can you grant my google account (bmenasha@google.com) access to the CAI export that fails to be imported into BigQuery? 
thanks

MarcFundenberger:
Done, you should have access to (both link to same file): 
https://storage.cloud.google.com/adeo-resource-inventory-dump-preprod/ExportAsset-2024-10-23.dump
gs://adeo-resource-inventory-dump-preprod/ExportAsset-2024-10-23.dump

MarcFundenberger:
From additional tests, it seems that the culprit is the "cloudbuild.googleapis.com/Build" asset type.
In a "full" resource dump, I have over 1.6 million of such assets.
AFAIK, there is not mention of such an asset type in the Cloud Asset Inventory documentation or release Notes...

bmenasha:
this pull request should resolve it. https://github.com/GoogleCloudPlatform/professional-services/pull/1394    
I'll update the latest pipeline after it's merged. 

thanks

bmenasha:
change is merged. dataflow template was updated. let me know if there are any problems. 
thanks

MarcFundenberger:
Thanks for the update.
I'll try a "full" dump including the problematic resource and I'll let you know how it went.

MarcFundenberger:
Allright, the problematic asset type is successfully loaded with the new pipeline version, thank you !

PR #1394, opened 2024-11-15
Address issue #1373 and issue #1374 and other performance/bug fixes

bmenasha:
Issue # 1374:
    Use the latest Dataflow SDK version.

Issue # 1373: Unable to deal with new cloudbuild.googleapis.com/Build assets:

    The core issue was that the discovery_name of this new asset type is
incorrectly reported as cloudbuild.googleapis.com/Build rather than 'Build'. Try to deal with that by correcting any discovery_name with a '/' in it. But there are other fixes necessary to speed processing.

Other performance/bug fixes:

- Use the discovery document generated schema if we have one over any resource generated one. This is a big performance improvement as determining the schema from the resource is time consuming, it's also not productive as if we have an API resource schema, it should always match the resource json anyways.

- Add ancestors, update_time, location, json_data to discovery generated schema. This prevents those properties from being dropped if we always rely on it.

- Sanitize discovery document generated schemas. If we are to always rely on them, it's possible they could be invalid, so enforce the bigquery rules on them as well.

- Use copy.deepcopy less, only when we copy a source into a destination field.

- Prevent bigquery columns with BQ_FORBIDDEN_PREFIXES from being created. There are some bigtable resources that can include these prefixes.

- Some BigQuery model resources had NaN and Infinity values for numeric fields. Try to handle those in sanitization.

- When merging schemas, stop after we have BQ_MAX_COLUMNS fields. This helps to stop the merge process earlier. (It can take forever if there are many unique fields and many elements).

- When enforcing schema on a resource, recognize when we are handling addition properties and add the additional property fields to the value of the additional property key value list in push_down_additional_properties. This produced more regular schemas.

- Add ignore_unknown_values to the load job so that we don't fail if resource contains fields not present in the schema.

- Accept and pass --add-load-date-suffix via main.py.

- Better naming of some local variables for readability.

- Some format changes suggested by Intellij.

## Pull Request Template
Please, go through these steps before you submit a PR.

1. Make sure that your PR is not a duplicate.
2. **Before** submitting the PR, review your changes:
    * Adjust the code to the existing [style](https://google.github.io/styleguide).
    * Add a set of unit tests for the changes and ensure they all pass.
    * Add accompanying README.md file with instructions on usage. See [awesome-readme](https://github.com/matiassingers/awesome-readme) for good examples of high-quality READMEs.
    * Add a link to your contribution in the top-level [README](https://github.com/GoogleCloudPlatform/professional-services/blob/master/README.md) (alpha-order).
    * Add [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0) license headers with an up-to-date copyright date attributed to `Google LLC`.
    * Remove unnecessary LICENSE files. There's no need to include an additional license since all repository submissions are covered by the top-level Apache 2.0 [license](https://github.com/GoogleCloudPlatform/professional-services/blob/master/LICENSE).

3. **After** these steps, you're ready to open a pull request.
    * Give a descriptive title to your PR.
    * Provide a description of your changes.
    * Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).
<br/>

**PLEASE REMOVE THIS TEMPLATE BEFORE SUBMITTING**


agold-rh (approved):

agold-rh:
@bmenasha There seems to be a problem pulling one of the trigger images. Last time this happened, it cleared up in an hour or so. I'll run the trigger again in a couple of hours; with luck, that will fix the problem.

PR #1430, opened 2025-01-23
Delete examples/dataflow-bigquery-transpose directory

svetakvsundhar:
This example is 8 years old and thus has some old dependencies on no longer up-to-date client libraries; eg

`Beam 2.9.0`
`bigquery 1.56`
`Java 8`.

I propose removing directory from our public git since it no longer compiles with recent client libraries.


rachelgagnon:
cool thanks for starting this! just for clear documentation, which recent versions of the client libraries did you try to compile with (or at least Beam version)?

rachelgagnon (commented):
cool thanks for starting this! just for clear documentation, which recent versions of the client libraries did you try to compile with (or at least Beam version)?

svetakvsundhar:
tried with `2.61.0`. but main criteria for this deletion is that the use case/example is very old.

rachelgagnon (approved):

ryanmcdowell (approved):
Approved. 

This solution additionally is no longer valid as the PIVOT statement exists natively in BigQuery:
https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#pivot_operator

PR #1432, opened 2025-01-23
Delete examples/dataflow-scala-kafka2avro directory

svetakvsundhar:
This example leverages [Scio](https://github.com/spotify/scio), but I don't think it's generalized enough to belong in this repo. A better home for it is potentially the [Scio cookbook](https://github.com/spotify/scio/tree/d5d20ade24d488af8af4a8c9e13742154cab7f44/scio-examples/src/main/java/org/apache/beam/examples/cookbook).


svetakvsundhar:
R: @ryanmcdowell 
R: @rachelgagnon

iht:
I oppose against removing this example. Why not having a Dataflow Scala example in this repo?

iht (approved):
I spoke with @svetakvsundhar, this example is probably now outdated.  Long live my little and beatiful example 😭

PR #1446, opened 2025-02-11
Setup service extensions(callout) with GCP Application Load Balancer (EXTERNAL_MANAGED)

pawanphalak:
### OVERVIEW:

[Callout-based Service Extensions](https://cloud.google.com/service-extensions/docs/overview) let users of Google Cloud products, such as Cloud Load Balancing and Media CDN, to insert programmability directly into the data path. This helps you customize the behavior of these proxies to meet your business needs.
Callouts let you use Cloud Load Balancing to make gRPC calls to user-managed services during data processing.
You write callouts against Envoy's external processing gRPC API (ext-proc). Callouts run as general-purpose gRPC servers on user-managed compute VMs and Google Kubernetes Engine Pods on Google Cloud, multicloud, or on-premises environments.

This repository contains the code to set up the callout-based service extension. The service extension is configured with a Global Application Load Balancer, and the service extension backend runs on GKE. This code repository also highlights detailed steps that are automated to provision a sample application, service extension backend (which modifies the specific request header), and deployment of all GCP resources automated with Terraform.

The service extension code in this repository follows a specific use case where a client header key `secret` value is Base64 encoded by the service extension and sent back to the GCP load balancer backend. This is a common use case where users need to modify headers with their own custom logic and send them back to the backend with the updated value.

This repository contains detailed steps to provision the resources along with the configuration steps in an automated way to setup the callout based service extension.

Service Extension

### Tests

Attaching unit tests results for the golang files:

Screenshot 2025-02-11 at 3 20 02 PM


agold-rh:
@pawanphalak Can you point me to other examples in this repo that include *.key, *.crt, *.pem files? I can't seem to find any others using the github search, and I'm hesitant to approve code that includes keys.

pawanphalak:
@agold-rh Here is the example from the official service extension repo which uses the same keys - https://github.com/GoogleCloudPlatform/service-extensions/tree/main/callouts/go/extproc/ssl_creds. End users can use there own keys, these example keys can serve as a reference to get started.

pawanphalak:
@agold-rh I have removed all the certificates and keys from this pull request and instead added a step in a readme to pull those directly from the official repository.

agold-rh:
@pawanphalak Some comments from other administrators:

"Tbh... It doesnt really matter if they're example keys, we shouldn't commit them. They should be placeholders for the customer to fill in."

"A compromise I’ve seen and used in the past is to include empty secrets.json.EXAMPLE (note extra suffix) artifacts to illustrate the import path while mitigating secret commits. This can be hardened with the .gitignore policies."

All in all, the consensus is to disallow the commit of any actual keys.

pawanphalak:
@agold-rh Thanks for sharing the feedback. I will make a note for the future commits. All the keys and certificate should be removed from this PR now. Please let me know if there are any other action items.

PR #2102, opened 2026-06-23
feat(agent-eval): add native ADK2 Workflow client execution and case-level granular filtering

svidiella:
This PR adds native ADK2 Workflow execution support to the in-process runner (under `BaseNode` / `NodeRunner`) and establishes a case-level granular filtering mechanism (`case_id`) to execute specific eval cases selectively.

### DIFF
--- a/tools/agent-eval/src/agent_eval/core/agent_client.py
+++ b/tools/agent-eval/src/agent_eval/core/agent_client.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 import contextlib
 import json
 import logging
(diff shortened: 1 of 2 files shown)


tools/agent-eval/src/agent_eval/core/agent_client.py
```diff
@@ -665,11 +669,19 @@ async def run_interaction(
...
         )
+        inv_context._event_queue = asyncio.Queue()
 
         # 2. Run the agent in-process!
         response_text = ""
 
-        if hasattr(self.agent_instance, "run_async"):
+        if isinstance(self.agent_instance, BaseNode):
+            root_ctx = AdkContext(inv_context)
+            node_runner = NodeRunner(node=self.agent_instance, parent_ctx=root_ctx)
```
lucieyxu: instead of directly using the hidden class, can we wrap the workflow in an app?

lucieyxu (approved):

PR #2103, opened 2026-06-23
fix(agent-eval): pass results_csv to generate_html_report in SDK wrapper

svidiella:
### Description
This PR fixes a critical bug in the `agent-eval` Python SDK wrapper where the evaluation results CSV path (`results_csv`) was not being passed to the HTML report generator (`generate_html_report`).

### Root Cause
In `agent_eval/sdk.py`, the `run_evaluation` function executes the evaluation, successfully resolves the raw results CSV file path as `latest_csv`, and loads it into a pandas DataFrame. However, when invoking `generate_html_report`, it only passed `run_dir` and `summary`, completely omitting the `results_csv` argument.

Because `generate_html_report` defaults `results_csv` to `None`, this caused the HTML report builder to silently skip parsing the CSV, resulting in an HTML report where the **conversation histories (turns)**, **tool calls**, and **state variables** were completely blank (empty lists and dicts) in the interactive per-question deep-dive accordions.

### Solution
Pass `results_csv=latest_csv` in the `generate_html_report` invocation inside `sdk.py`. This fully restores the interactive deep-dive sections in the generated `report.html` file when running evaluations programmatically via the Python SDK.

### Testing
- Verified locally by running programmatic evaluations inside a FastAPI orchestrator test suite.
- Re-generated the HTML report using the corrected SDK code, confirming that conversation histories, tool calls, and state variables now render beautifully in the UI.

### DIFF
--- a/tools/agent-eval/src/agent_eval/core/simulation.py
+++ b/tools/agent-eval/src/agent_eval/core/simulation.py
@@ -144,6 +144,8 @@ class PrePopulatingSessionService(InMemorySessionService):
             parts = session_id[len(EVAL_SESSION_ID_PREFIX) :].split("___")
             if len(parts) > 0:
                 case_id = parts[0]
+                if "_rep_" in case_id:
+                    case_id = case_id.split("_rep_")[0]
 
         session = await super().create_session(
             app_name=app_name, user_id=user_id, state=state, session_id=session_id
@@ -204,6 +206,7 @@ async def run_simulation_in_process(
     run_mode: str = "all",
     case_id: str | None = None,
     agent_instance: Any = None,
+    replications: int = 1,
 ) -> list[dict[str, Any]]:
     """Runs the simulation in-process using ADK Python APIs and returns converted records."""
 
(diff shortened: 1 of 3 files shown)


lucieyxu (dismissed):

tools/agent-eval/src/agent_eval/core/simulation.py
```diff
@@ -144,6 +144,8 @@ async def create_session(self, app_name, user_id, state, session_id):
             parts = session_id[len(EVAL_SESSION_ID_PREFIX) :].split("___")
             if len(parts) > 0:
                 case_id = parts[0]
+                if "_rep_" in case_id:
+                    case_id = case_id.split("_rep_")[0]
```
lucieyxu, line 148: Is this replication automatically generated? what is it exactly doing? Maybe worth explaining in the docstring or in a comment

tools/agent-eval/src/agent_eval/core/simulation.py
```diff
@@ -204,6 +206,7 @@ async def run_simulation_in_process(
     run_mode: str = "all",
     case_id: str | None = None,
     agent_instance: Any = None,
+    replications: int = 1,
 ) -> list[dict[str, Any]]:
     """Runs the simulation in-process using ADK Python APIs and returns converted records."""
```
lucieyxu, line 211: Mayve here, explanation on replications

svidiella: updated the docs

tools/agent-eval/src/agent_eval/core/simulation.py
svidiella: check if enough or need more explanation

lucieyxu (approved):

PR #2107, opened 2026-06-26
fix( agent-eval cli): restrict load_dotenv to local .env to prevent parent folder…

hselbie:
## Description
This PR resolves an issue in `agent-eval setup` and `agent-eval init` where the active Google Cloud project ID was silently overridden by a parent directory's `.env` file.

### Root Cause
By default, calling `dotenv.load_dotenv(override=True)` without an explicit path searches recursively upwards
  through parent directories. If any parent directory contains a `.env` file (e.g., in nested developer workspace
  structures), those variables are loaded and silently override shell environment variables or local `gcloud`
  configuration defaults.

### Changes
- **CLI Fix**: Updated `load_dotenv` calls in `src/agent_eval/cli/commands/setup.py` and
  `src/agent_eval/cli/commands/init.py` to specify `dotenv_path=".env"`. This restricts the search scope strictly to
  the local directory's `.env` file.
- **Test Isolation**: Added an autouse pytest fixture to `tests/core/test_config.py` to patch `EvalConfig.
  model_config` with `env_file=None` so unit tests are completely isolated from local `.env` files.
- **SDK Test Alignment**: Updated `tests/test_sdk.py` mock assertions to include default parameters
  `case_id=None` and `replications=1`.

### DIFF
--- a/tools/agent-eval/src/agent_eval/cli/commands/init.py
+++ b/tools/agent-eval/src/agent_eval/cli/commands/init.py
@@ -811,7 +811,7 @@ def _verify_environment(auto_approve: bool = False) -> None:
         _check_api_enabled as _setup_check_api,
     )
 
-    load_dotenv(override=True)
+    load_dotenv(dotenv_path=".env", override=True)
 
     console.print()
     console.print(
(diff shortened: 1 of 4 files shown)


svidiella (approved):
LGTM

